PR #1, opened 2025-06-13
Analytics via Automattic Tracks

etoledom:
Closes GRA-185

## Description

This PR implements Automattic Tracks for analytics

I've put the effort on making the interface type-safe, and avoiding gigantic enum declarations.

`Analytics` is added to `EnvironmentValues` for easy access.

The usage looks like this:

```swift
analytics.track(WelcomeScreenEvent.authButtonPressed)
```

And the events declaration looks like this:

```swift
// One events file per screen (or app section), wrapping all related event in an enum for easy access
enum WelcomeScreenEvent {
    static let authButtonPressed: AnalyticsEvent = AuthButtonPressed()
}

// The event struct is a bit more verbose, but this allows easy type safety and avoid declaring property keys with strings.
// Declared as private to avoid populating the project with types.
private struct AuthButtonPressed: AnalyticsEvent {
    let name: String = "auth_button_pressed"
    let properties: EventProperties? = nil
}
```

An example of an event with properties:

Usage:
```swift
analytics.track(WelcomeScreenEvent.authFailed(with: error.localizedDescription))
```

Declaration:

```swift
enum WelcomeScreenEvent {
    static func authFailed(with error: String) -> AnalyticsEvent {
        AuthFailed(properties: AuthFailed.Properties(error: error))
    }
}

private struct AuthFailed: AnalyticsEvent {
    struct Properties: Encodable, Sendable {
        let error: String
    }

    var name: String = "auth_failed"
    var properties: EventProperties?
}
```

I intentionally avoided using generic type for `AnalyticsEvent` (generic on the properties type) to make it easy to declare properties as nil. Though it's probably not a big deal 🤔 

## Test:

**Note:** The event names and the actual current interaction on the app doesn't match. This is made for tests purposes only.

- Run the app
- Fetch the profile from an email
- Go to Tracks and search for the events 
  - Can search by event name: `gravatar_ios%` 
  - They might take 6~10 mins to appear
  - Send the app to background and foreground to force sending the events to tracks.
- Alternatively, check the tracks request with ProxyMate or similar

### DIFF
--- a/GravatarApp.xcodeproj/project.pbxproj
+++ b/GravatarApp.xcodeproj/project.pbxproj
@@ -360,7 +360,8 @@
 				SDKROOT = auto;
 				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator macosx xros xrsimulator";
 				SWIFT_EMIT_LOC_STRINGS = YES;
-				SWIFT_VERSION = 5.0;
+				SWIFT_STRICT_MEMORY_SAFETY = NO;
+				SWIFT_VERSION = 6.0;
 				TARGETED_DEVICE_FAMILY = "1,2,7";
 				XROS_DEPLOYMENT_TARGET = 2.0;
 			};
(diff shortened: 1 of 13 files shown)


pinarol (approved):
LGTM 🎉

PR #2, opened 2025-06-16
Swiftlint

etoledom:
This PR implements the same SwiftFormat system used in the SDK, with the same set of rules.

It uses SwiftFormat via Docker, and it's run from the `Makefile`.

`$ make lint` - Will fail if there are syntax error, but won't modify the sources.
`$ make swiftformat` - Will modify the sources with the appropriate syntax.

**Note:** I'd like to rename `swiftformat` to `format` for simplicity, but I'm worried we could get confused between SDK and App development.

**Note 2:** As we introduce the `Makefile`, now also is available the `make dev` command to open the project.

## Test:

- Make a known syntax error on the source
- Run `make swiftformat`
  - Check that the syntax error has been fixed.


pinarol (approved):
LGTM 🎉

PR #3, opened 2025-06-17
Add basic Buildkite pipeline for linters

wzieba:
Closes AINFRA-768

### Description

This PR introduces a basic CI pipeline that runs `Swiftformat` and tests of the project.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -14,7 +14,7 @@ steps:
     plugins:
       - docker#v5.12.0:
           image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
-          command: [ "Gravatar", "GravatarAppTests", "Packages", "--lint" ]
+          command: [ "GravatarApp", "GravatarAppTests", "Packages", "--lint" ]
           workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
     notify:
       - github_commit_status:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,8 +5,6 @@
 
 agents:
   queue: "mac"
-env:
-  IMAGE_ID: $IMAGE_ID
 
 steps:
   - label: ":swift: SwiftFormat Linting"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,32 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+# Variables used in this pipeline are defined in `shared-pipeline-vars`, which is `source`'d before calling `buidkite-agent pipeline upload`
+
+agents:
+  queue: "mac"
+env:
+  IMAGE_ID: $IMAGE_ID
+
+steps:
+  - label: ":swift: SwiftLint"
+    command: swiftlint
+    key: swiftlint
+    notify:
+      - github_commit_status:
+          context: "SwiftLint"
+    agents:
+      queue: "linter"
+
+  - label: ":swift: SwiftFormat Linting"
+    key: swiftformat-lint
+    plugins:
+      - docker#v5.12.0:
+          image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+          command: [ "Sources", "Tests", "--lint" ]
+          workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
+    notify:
+      - github_commit_status:
+          context: "SwiftFormat Linting"
+    agents:
+      queue: "default"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,5 +19,11 @@ steps:
     notify:
       - github_commit_status:
           context: "SwiftFormat Linting"
-    agents:
-      queue: "default"
+
+  - label: "🧪 Tests"
+    key: tests
+    command: |
+      xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16'
+    notify:
+      - github_commit_status:
+          context: "Unit tests"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,6 +19,8 @@ steps:
     notify:
       - github_commit_status:
           context: "SwiftFormat Linting"
+    agents:
+      queue: "default"
 
   - label: "🧪 Tests"
     key: tests
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -23,7 +23,7 @@ steps:
   - label: "🧪 Tests"
     key: tests
     command: |
-      xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16'
+      xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16,arch=arm64'
     notify:
       - github_commit_status:
           context: "Unit tests"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,6 +5,8 @@
 
 agents:
   queue: "mac"
+env:
+  IMAGE_ID: $IMAGE_ID
 
 steps:
   - label: ":swift: SwiftFormat Linting"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,21 +9,12 @@ env:
   IMAGE_ID: $IMAGE_ID
 
 steps:
-  - label: ":swift: SwiftLint"
-    command: swiftlint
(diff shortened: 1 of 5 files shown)


.buildkite/shared-pipeline-vars
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# This file is `source`'d before calling `buildkite-agent pipeline upload`, and can be used
+# to set up some variables that will be interpolated in the `.yml` pipeline before uploading it.
+
+export IMAGE_ID="xcode-26.0-beta"
```
wzieba: I think `26.0-beta` is fine for this project: it's not meant to be released very soon, but the project can already align with the newest tooling.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,31 @@
...
+    notify:
+      - github_commit_status:
+          context: "SwiftFormat Linting"
+    agents:
+      queue: "default"
+
+  - label: "🧪 Tests"
+    key: tests
+    command: |
+      xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16,arch=arm64'
```
wzieba, line 28: I understand it's rather not typical way of running tests comparing to other projects, but in scope of this PR I didn't want to introduce the whole Fastlane setup, as it requires more work and thinking.

I believe running lint + tests are good enough for having an initial CI setup that we can improve over time.

.buildkite/shared-pipeline-vars
AliSoftware: I'd tend to agree, but I'd also let this kind of decision to be validate by the dev team first just to be sure :) /cc @pinarol @andrewdmontgomery

AliSoftware (approved):
Such an initial setup looks good to me as a starting point.

The team already mentioned they are ok to take care of evolving the pipeline as they work on the project and add a `Fastfile` later and get more needs for what to run in CI etc. so I think this will evolve anyway.

(Added a couple of members of the Gravatar iOS team as reviewers still just to get their eyes on it)

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,31 @@
...
+  IMAGE_ID: $IMAGE_ID
+
+steps:
+  - label: ":swift: SwiftFormat Linting"
+    key: swiftformat-lint
+    plugins:
+      - docker#v5.12.0:
+          image: "ghcr.io/nicklockwood/swiftformat:$SWIFTFORMAT_VERSION"
+          command: [ "GravatarApp", "GravatarAppTests", "Packages", "--lint" ]
+          workdir: "${BUILDKITE_BUILD_CHECKOUT_PATH}"
```
AliSoftware, line 18: 💭  Note that on one end I like that we run this in `docker` and leverage the existence of the dedicated swiftformat docker image 👍 ❤️  but otoh I realized this might need to be kept in sync with how it's been introduced via `make lint` in https://github.com/Automattic/Gravatar-iOS/pull/2, so maybe someone from the Gravatar dev team can chime in there to share their thoughts on how they plan to use Swiftlint and how we could keep the configs in sync? /cc @etoledom @pinarol

.buildkite/pipeline.yml
etoledom: > but in scope of this PR I didn't want to introduce the whole Fastlane setup, as it requires more work and thinking.

As long as we have tests running on CI, sounds great to me. @andrewdmontgomery - any thought?

> I believe running lint + tests are good enough for having an initial CI setup that we can improve over time.

Sounds good to me!

Just to note that at some point (soon~ish) we will need to have a testable build on PRs for internal testing. We are hoping to have a simple released version of this app for mid August.

andrewdmontgomery: > but in scope of this PR I didn't want to introduce the whole Fastlane setup, as it requires more work and thinking.

Fastlane is definitely in our future for this repo.  But I love iterating.  This looks great.

etoledom: Question: Will this also run the tests from the local swift packages?

.buildkite/pipeline.yml
etoledom: Basically we have decided to continue having the same workflow than in the SDK, so we have copied the setup.
We use `make swiftformat` manually when we want to auto-fix the syntax. And we decided to not have inline warnings as they can be distracted.

> maybe someone from the Gravatar dev team can chime in there to share their thoughts on how they plan to use Swiftlint and how we could keep the configs in sync?

We survived with this setup on the SDK, so it shouldn't be much of a problem for us now.
I guess we can't run the `make` command from .buildkite pipeline directly?

Ultimately we can add a reminder comment on the make file to update `.buildkite/pipeline.yml` if we ever modify the `make swiftformat` command 🤔

.buildkite/pipeline.yml
etoledom: If we use a custom test plan on GravatarApp target which includes all packages we can run them all with this same command.

I don't know what are the best practices on CI though, but this seems to me good in a way that we don't need to modify buildkit pipelines for every new package we create 🤔

.buildkite/shared-pipeline-vars
pinarol: Thanks for the ping! Our v1.0 deadline is really tight so not sure the stable version will be out by then. For that reason I think it'll be safer to stick to XCode 16.4 for now.

andrewdmontgomery:
It's a good callout.  We are duplicating the docker configuration, having it both in the Makefile and in the pipeline definition.

For the SDK, I haven't been able to think of a way to take the existing approach and deduplicate this part.

But since this is an Xcode project, @AliSoftware's question "how they plan to use Swiftlint" points us in another direction. It might be more useful for us to implement it as an [Xcode build phase](https://github.com/nicklockwood/SwiftFormat?tab=readme-ov-file#xcode-build-phase).

On the upside:

* We would have one SwiftFormat config/instantiation for development and CI
* We could still pin the version of SwiftFormat
* And of course, Builds in Xcode would automatically apply formatting 🎉 

On the downside:

* Builds in CI will fail if there are SwiftFormat problems (since it will run during the build phase), which isn't ideal in terms of CI output
  * We might still be able to include a dedicated `--lint` check, although I'd need to try that first

I'll try this in a bit and see how it works.

etoledom:
> It might be more useful for us to implement it as an [Xcode build phase](https://github.com/nicklockwood/SwiftFormat?tab=readme-ov-file#xcode-build-phase).

@andrewdmontgomery - I experimented with this, and found out a few more downsides:

- Changing format on every build might be unexpected.
- Losing undo history when there's an automatic formatting.
- Inline warnings can be distracting.
- Possible build issues as this is integrated with the build phase, and (probably quite minor) increased build times.

I tried to find a way to manually trigger the swift format task from within xcode but didn't find one.
I didn't find a way to run `--lint` version of the command from the build plugin to avoid auto-formatting, though breaking the build because of formatting could be even more annoying 🤔  

And the command plugin is only available for packages.

The only reason I wanted to have this more automated was to not forget to run the format before pushing to remote, but at the end I like this approach where the swift syntax formatter is works around the project and is not in it.

That said, I'm not completely against using the build plugin, I just slightly prefer this less intrusive approach.

.buildkite/shared-pipeline-vars
wzieba: Sure 👍 addressed via 37bad2449df386dcc29ee9cc06f30543f37f0681

.buildkite/pipeline.yml
wzieba: Thanks, I don't think the tests from local swift packages are executed now. From what I understand we can either:
- create a custom test plan and include all targets 
- add targets to the existing scheme (`Test` section) and run tests as it's now declared in `pipeline.yml`

@AliSoftware do you have any preference?

andrewdmontgomery:
Gotcha.  Then let's iterate.

The built-in `swift-format` option in Xcode is only invokable manually, and only on a specific file.  When I discovered that a few months ago, I considered that a limitation.

But given your observations about being surprised by build-time format changes, maybe that's a better pattern after all.

andrewdmontgomery:
I don't mean to say that we should use the built in `swift-format` by the way.  Obviously we can consider it.  I only meant that the "manual invocation", and "per file" aspects are more interesting to me now.

.buildkite/pipeline.yml
AliSoftware: Personally I think test plans are a better fit, but at that point I'll let the Gravatar team decide as they're the ones who will work with those and have the best idea how many local packages they plan to add over time

wzieba: ok, I'll prepare the test plan 👍

wzieba: All tests were executed (0 from main module, 5 from Analytics): https://buildkite.com/automattic/gravatar-ios/builds/13/steps/canvas?jid=01977e8d-f2a3-43a2-b0b3-5c87893f59cb#01977e8d-f2a3-43a2-b0b3-5c87893f59cb/222-3778

wzieba:
Thanks all for discussions, I think I addressed all action items - if I missed something, please ping me!

.buildkite/pipeline.yml
andrewdmontgomery: +1 for a test plan, thanks!

PR #5, opened 2025-06-18
Analytics: Update userID key and anonymous user ID

etoledom:
Closes GRA-202

This PR updates the user Key to be the same than Gravatar web.
It also introduces anonymous user ID, to be used for users whom haven't been authenticated yes.

This anonymous user ID will persist through the life of the app instalation.

## Test:

- Logout if needed
- Login to your account
- Check Tracks live view events filtering by event name `gravatarios%`
  - First `auth_button_pressed` event should have `useridtype`: `anon` and an UUID userID
  - `auth_success` and subsequent `auth_button_pressed` should have:
    -  `useridtype`: `wpcom:userid`, plus the 
    -  `userid` - set to a number (wpcom user id)
    - User column should be filled with the user name


pinarol (commented):
> First auth_button_pressed event should have user Key: anon and an UUID userID

WDYM user Key? I see a value as `anon` but it has the key `useridtype`.

> auth_success and subsequent auth_button_pressed should have userIDKey: wpcom:userid

I don't see `auth_success`. I only see `auth_button_pressed`. Which is odd.

etoledom:
> WDYM user Key? I see a value as anon but it has the key useridtype.

Yes, my mistake. I meant `useridtype`

> I don't see auth_success. I only see auth_button_pressed. Which is odd.

This was fixed with a PR on Tracks backend side.

*Note:* There's a known issue of multiple events being sent to the backend. 

@pinarol - This is ready to take a look 👀  I have updated the test steps

pinarol (approved):
Looks good and works as described 👍

PR #6, opened 2025-06-18
OAuth

etoledom:
Closes: GRA-187

OAuth for Gravatar App.

This is based on the OAuth code from the SDK, but trying to make it more organized, simpler, and stand-alone.

We want to use the code type of authorization to get the access token, but so far it's been failing for me, so on this PR we are using the token type. I still added the code parser for a fast switch in the future.

There's a new make command to create the secrets file. `make setup` it should be called before the project is first built.
I have linked it to `make dev` to avoid running 2 commands at once.

There's some basic UI with example usage, but don't mind much about the App side of the code (apart from the public API and usage of the OAuth package). Those views are going to change completely in the future.

### How to test:

- Run `make setup` or `make dev`.
- Open the newly created `Secrets/Secrets.swift` file and fill it with the apps secrets.
- Run the app
- Tap on the login button
- Complete the oauth flow on web
  - Check that the logged-in screen appears with the user display name.
- Close and restart the app
  - Check that the app persisted the logged in state
- Press log-out
  - Check that you are back to the welcome screen.


pinarol (approved):
Looks good and works as described. 👍

PR #7, opened 2025-06-19
Configure code signing and add tooling to manage it

mokagio:
See https://linear.app/a8c/issue/AINFRA-786 and https://linear.app/a8c/issue/AINFRA-792

This PR:

- Add Ruby tooling, in particular Fastlane, to manage the distribution provisioning profile
- Establishes `xcconfig` files where to store DRY configurations and uses it do define the code signing settings

Also:

- Removes the macOS destination from the project, [as the SDK does not support macOS at this time](https://github.com/Automattic/Gravatar-SDK-iOS/blob/71350ef7274a989ca2ed89c512f8b9dd99bc8c3d/Package.swift#L12).
- Tidies up the test plan, which had a duplicated reference in the project and would have been bundled with the app as a resource

What the code does not show are the counterpart settings in the Developer Portal and App Store Connect, where `com.gravatar.GravatarApp` was created and configured, _including having Associated Domains entitlement_

## Testing

Checkout the branch and verify the app builds and runs on device.

As an Automattic developer, you should be able to download the development profile via our internal management system. Or, if you're logged into Xcode with the Automattic Apple ID, Xcode should offer to download it for you.


fastlane/lib/env_manager.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'dotenv'
+# TODO: It would be nice to decouple this from Fastlane.
+# To give a good UX in the current use case, however, it's best to access the Fastlane UI methods directly.
+require 'fastlane'
+
+# Manages loading of environment variables from a .env and accessing them in a user-friendly way.
+class EnvManager
```
mokagio, line 9: This will eventually be part of release toolkit and be removed from this repo — https://linear.app/a8c/issue/AINFRA-242

pinarol (approved):
It works 🎉 Thank you 🙇 

(etoledom is AFK)

AliSoftware (approved):

PR #8, opened 2025-06-19
Cache Swift packages in CI

mokagio:
This should make the build a bit faster.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -25,7 +25,7 @@ steps:
   - label: "🧪 Tests"
     key: tests
     command: |
-      xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16 Pro'
+      make test
     notify:
       - github_commit_status:
           context: "Unit tests"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -24,7 +24,9 @@ steps:
 
   - label: "🧪 Tests"
     key: tests
+    plugins: [$CI_TOOLKIT]
     command: |
+      install_swiftpm_dependencies
       make test
     notify:
       - github_commit_status:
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -24,8 +24,10 @@ steps:
 
   - label: "🧪 Tests"
     key: tests
+    plugins: [$CI_TOOLKIT]
     command: |
-      xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16,arch=arm64'
+      install_swiftpm_dependencies
+      make test
```
mokagio, line 30: The move from calling `xcodebuild` here to a `make` task is so that if a dev wants to run the same command as CI does locally, they can do so.

Makefile
```diff
@@ -24,6 +24,9 @@ help:  # Display this help.
 dev: # Open the package in xcode
 	xed .
 
+test: # Run the app unit tests
+	@xcodebuild test -project GravatarApp.xcodeproj -scheme GravatarApp -destination 'platform=iOS Simulator,name=iPhone 16 Pro'
```
mokagio, line 28: Notice this now uses iPhone 16 Pro. This is a bit of an arbitrary change, but I noticed that if I run the tests on my machine with iPhone 16, the Simulator fails to launch:

```
Test target AnalyticsTests encountered an error (Failed to clone device named 'iPhone 16'. (Underlying Error: The operation couldn’t be completed. Device was allocated but was stuck in creation state.  Check CoreSimulator.log for more information.))
```

I tried to run the tests via Fastlane's `scan` and it automatically picked up iPhone 16 Pro and succeeded to run the tests.

When I changed the command to use iPhone 16 Pro on my end via vanilla `xcodebuild` like CI does, it worked.

etoledom:
@mokagio - I think this looks good. It's Ok to use iPhone 16 Pro.

Is this PR ready? Any reason that it is still a draft?
Thank you 🙏

PR #9, opened 2025-06-20
Package and Distribute Prototype Build from CI

mokagio:
See https://linear.app/a8c/issue/AINFRA-808

Distribution works as proven by the comment below:

image


Config/GravatarApp.debug.xcconfig
```diff
@@ -7,4 +7,3 @@
 CODE_SIGN_IDENTITY = Apple Development: Created via API (886NX39KP6)
 
 PROVISIONING_PROFILE_SPECIFIER = Gravatar iOS Development
-PROVISIONING_PROFILE_SPECIFIER[sdk=macosx*] = Gravatar macOS Development
```
mokagio, line 10: This was a leftover from an experiment I made in #7

.buildkite/commands/setup_shared.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -eu
+
+echo "--- :ruby: Setting up Ruby tools"
+install_gems
```
mokagio, line 4: Decided to split between "shared" and "for building app" because, as you can see, they represent different steps. I trust that as the app grows in need, we'll make good use of the separation. But if nothing else, this saves us from duplicating the `echo "--- :ruby: Setting up Ruby tools"; install_gems` calls :)

.buildkite/pipeline.yml
```diff
@@ -29,3 +29,24 @@ steps:
...
+          - github_commit_status:
+              context: Package Prototype Build
+
+      - label: Distribute Prototype Build
+        depends_on: prototype_build
+        command: .buildkite/commands/distribute_prototype.sh
+        plugins: [$CI_TOOLKIT]
+        notify:
+          - github_commit_status:
+              context: Distribute Prototype Build
```
mokagio, line 53: Open to removing the `group` node. The new canvas makes the logical connection already clear:

image

vs

image

GravatarApp.xcodeproj/project.pbxproj
```diff
@@ -455,6 +447,118 @@
 			};
 			name = Release;
 		};
+		3FF1E3952E053CEF0082DA05 /* Enterprise */ = {
```
mokagio, line 468: At this point in the project, the Enterprise/Prototype-Build behaves in the same way as the Release build.

I thought I could get away without creating a dedicated build configuration and passing the build settings for code signing to Fastlane.

However, when I tried that, I couldn't get `xcodebuild` to archive because the process applied the build settings to the Swift packages it built, but those do not support provisioning profiles. Example:

```
[16:43:57]: ▸ ❌ /Users/gio/Developer/a8c/Gravatar-iOS/DerivedData/GravatarApp/SourcePackages/checkouts/Automattic-Tracks-iOS/Package.swift: AutomatticTracksiOS_AutomatticTracksModelObjC does not support provisioning profiles.
AutomatticTracksiOS_AutomatticTracksModelObjC does not support provisioning profiles, but provisioning profile match InHouse com.gravatar.GravatarApp.prototype has been manually specified. Set the provisioning profile value to "Automatic" in the build settings editor.
(in target 'AutomatticTracksiOS_AutomatticTracksModelObjC' from project 'AutomatticTracksiOS')
```

So, it's necessary to have a dedicated build configuration so that the code signing settings are applied only to the app target.

fastlane/Fastfile
```diff
@@ -44,26 +57,111 @@ end
 default_platform :ios
 
 platform :ios do
+  # Builds the app binary for Enterprise distribution (Prototype Build)
+  #
+  # @param fetch_code_signing [Boolean] If true, will call `configure_code_signing_enterprise` before building. Recommended on CI.
+  # @param build_number [String] The build number to use for the build. When in CI, defaults to the value of `BUILDKITE_BUILD_NUMBER` environment variable.
+  # @param version [String] The version to use for the build. When in CI, defaults to the PR number, falling back to the branch name.
+  #
+  lane :build_prototype do |fetch_code_signing: true, build_number: build_number_from_ci, version: pr_number_or_branch_name_from_ci|
```
mokagio, line 65: Maybe it's just me, but I like how this implementation does not have Buildkite hardcoded into it. By requiring `build_number` and `version` as call parameters, and by feeding them via dedicated methods (`build_number_from_ci` and `pr_number_or_branch_name_from_ci`) I feel the code is more separated, in particular the difference between CI the concept and Buildkite the provider.

fastlane/Fastfile
```diff
@@ -44,26 +57,111 @@ end
...
+      configure_code_signing_enterprise
+    else
+      UI.message('Skipping fetching certificates and provisioning profiles as requested.')
+    end
+
+    build_app(
+      scheme: 'GravatarApp',
+      configuration: 'Enterprise',
+      # Explicitly set iOS destination to avoid xcodebuild accidentallypicking up tvOS
+      destination: 'generic/platform=iOS',
```
mokagio: In #7, I removed macOS as a supported architecture.

I considered removing tvOS as well because I don't think there are plans at the moment to support it. But tvOS comes from free, and this so far is the only extra line of code it requires to coexist in the project. So, I decided to leave it.

However, happy to remove it in the name of YANGI. Let me know.

.buildkite/commands/distribute_prototype.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -eu
+
+"$(dirname "${BASH_SOURCE[0]}")/setup_shared.sh"
+
+echo "--- :arrow_down: Downloading Prototype Build"
+buildkite-agent artifact download ".build/artifacts/*.ipa" . --step prototype_build
+buildkite-agent artifact download ".build/artifacts/*.app.dSYM.zip" . --step prototype_build
+
+echo "--- :firebase: Distributing Prototype Build"
+bundle exec fastlane distribute_prototype
```
mokagio, line 10: Most, if not all, our previous implementations use the verb "upload". For some reason, this time around the verb "distribute" felt more appropriate.

- Pro: The automation uploads the build and makes it immediately available. Using upload would tell only half of the picture.
- Con: The fact that the distribution is automatic is not really something that we control from this side of the code, so maybe the verb "distribute" is not accurate.

I don't think it's a big deal either way. At this point in time, I prefer distributed. But happy to revert if you feel strongly against it or if we want to be consistent and consider renaming to distribute/something-else an action that should be take in one go across all repos.

.buildkite/pipeline.yml
AliSoftware: I like the group, let's keep it

fastlane/Fastfile
```diff
@@ -22,7 +22,19 @@ CODE_SIGNING_ENV_VARS = %w[
...
+ENTERPRISE_XCCONFIG_PATH = File.join(PROJECT_ROOT_FOLDER, 'Config', 'GravatarApp.enterprise.xcconfig')
+ENTERPRISE_XCCONFIG = Xcodeproj::Config.new(ENTERPRISE_XCCONFIG_PATH)
+
+ARTIFACTS_FOLDER = File.join(PROJECT_ROOT_FOLDER, '.build', 'artifacts')
+
+PROTOTYPE_BUILD_NAME = 'GravatarApp-Prototype'
+
+FIREBASE_APP_ID = '1:124902176124:ios:7bffd4c3c7143e3a1620f9'
+# For convenience, this is the same tester group as the SDK demo app.
+FIREBASE_TESTERS_GROUP = 'gravatarsdk-ios---prototype-builds'
```
AliSoftware: 👍

fastlane/Fastfile
```diff
@@ -44,26 +57,111 @@ end
...
+    if fetch_code_signing
+      configure_code_signing_enterprise
+    else
+      UI.message('Skipping fetching certificates and provisioning profiles as requested.')
+    end
+
+    build_app(
+      scheme: 'GravatarApp',
+      configuration: 'Enterprise',
+      # Explicitly set iOS destination to avoid xcodebuild accidentallypicking up tvOS
```
AliSoftware:
```suggestion
      # Explicitly set iOS destination to avoid xcodebuild accidentally picking up tvOS
```

fastlane/Fastfile
```diff
@@ -44,26 +57,111 @@ end
...
-    api_key = app_store_connect_api_key
+    case type
+    when 'appstore'
+      # When readonly = false and type = appstore, we need the API key instantiated to connect with App Store Connect.
+      EnvManager.require_env_vars!(*ASC_API_KEY_ENV_VARS)
+      api_key = app_store_connect_api_key
+    when 'enterprise'
+      # The Enterprise account APIs do not support authentication via API key.
+      # If we want to modify data (readonly = false) we need to authenticate manually.
+      prompt_user_for_app_store_connect_credentials
```
AliSoftware, line 153: 💡 ❓  With the latest announcements / improvements in ASC (announcement of the API for Enterprise in ≈ Sept 2024 iirc, WWDC…) I wonder if this is still true nowadays.

See also https://github.com/fastlane/fastlane/pull/22215 & https://github.com/fastlane/fastlane/pull/27460

So maybe worth checking this assumption again at some point?

fastlane/Fastfile
```diff
@@ -73,14 +171,44 @@ def asc_team_id!
...
+
+def commit_hash_from_ci
+  ENV.fetch('BUILDKITE_COMMIT', nil)
+end
+
+def pull_request_number
+  # Buildkite sets this env var to the PR number if on a PR, but to 'false' (and not nil) if not on a PR
+  pr_num = ENV.fetch('BUILDKITE_PULL_REQUEST', 'false')
+  pr_num == 'false' ? nil : Integer(pr_num)
+end
```
AliSoftware, line 213: 💡 Since we're doing this kind of dance (especially the `pull_request_number` helper to check for `false` and the `pr_number_or_branch_name_from_ci` to do the `&.then` etc…) I'd love for those helpers to be part of `EnvManager` at some point, especially when we'll finally move `EnvManager` to the `release-toolkit` one day, to avoid duplicating those helpers in all repos.

.buildkite/commands/distribute_prototype.sh
AliSoftware: I don't mind either name.

Arguably I'd prefer for us to keep things consistent… but we could do that the other way around (i.e. have all other repos adopt the "Distribute" wording from now on) if that really bothers us.

I think the history of the step being named "Prototype Upload" was because the fastlane action behind that step was `appcenter_upload`. But now that the action is `firebase_app_distribution`, that makes sense that "Prototype Distribution" feels more appropriate 😉

pinarol (approved):
Thank you! I confirm I can install the build. 🎉

fastlane/Fastfile
AliSoftware: Actually, in retrospect and after discussing about this with @pinarol in Slack (ref: p1750674026771499/1750670384.067809-slack-CC7L49W13), given:
 - The MC tool that a12s use to invite themselves shows the name of the "Testing Group"
 - This means that MC still shows "GravatarSDK iOS - Prototype Builds" for the single testing group listed in MC that is now used for both the SDK Demo and the App; which is confusing when one wants to invite themselves and can't find the App in the list
 - The Firebase Console doesn't seem to allow changing that Testing Group name once it's created, so we can't change that to "Gravatar iOS SDK & App - Prototype Builds" or something to indicate that Testing Group would be for both of them

So we could:
1. Either delete the `gravatarsdk-ios---prototype-builds` Testing Group and re-create it anew, but with a name that now indicates that it'd be used for both the SDK Demo _and_ the App. Which would also require everyone already in the existing Testing Group to re-invite themselves… not idea
2. Or just end up create a separate Testing Group for the Gravatar App, so that in MC they appear as 2 separate entries in the list, with descriptive-enough names.

I clearly thing the 2nd option is the easiest and the one that makes more sense (only listed option 1 for completeness tbh).

fastlane/Fastfile
mokagio: Definitely!

Added https://linear.app/a8c/issue/AINFRA-828 for this.

fastlane/Fastfile
mokagio: Great idea!

And having them there would also help us in the remote chance we'll change CI provider. `EnvManager` could check the CI-provider-specific env vars to determine where it's running and read the commit/branch/PR info accordingly.

mokagio: https://linear.app/a8c/issue/AINFRA-829

fastlane/Fastfile
mokagio: Agreed. Went for a dedicated group

image

mokagio:
@AliSoftware @pinarol I updated the PR with the new dedicated group. I'd appreciate if you could confirm you can download it. FWIW, it works on my end but it crashes on launch. The latter is an issue I have with all FAB apps apparently, which is why I'm asking for a second pair of eyes. Thanks!

pinarol:
> @AliSoftware @pinarol I updated the PR with the new dedicated group. I'd appreciate if you could confirm you can download it. 

Thank you! I can download it 🎉

AliSoftware:
> @AliSoftware @pinarol I updated the PR with the new dedicated group. I'd appreciate if you could confirm you can download it. FWIW, it works on my end but it crashes on launch. The latter is an issue I have with all FAB apps apparently, which is why I'm asking for a second pair of eyes. Thanks!

I can confirm that I was able to download and launch the latest Prototype Build `PR #9 (42)` that was distributed to the new "Gravatar iOS - Prototype Build" Testing group.

image

I also do see the new testing group in the self-serve MC tool 👍 
image

## Side notes

### I was already a tester in MC

I didn't have to use the MC tool to add myself to the new Testing Group, since apparently you already added all the members of the "Gravatar SDK iOS - Prototype Build" testing group to the new "Gravatar iOS - Prototype Build" testing group
 - I'm assuming that was intentional, so that those who already invited themselves to test the SDK demo app would automatically get invites for this iOS app without having to invite themselves to it manually. This makes sense to me, but just wanted to mentioned it to explain why, when I went to the MC page, I was already marked as being member of this testing group by then.
 - That being said, when going to the Firebase Console to compare the 2 testing groups, it seems that you invited _almost_ everyone who was part of the SDK testing group to the app testing group… except @pinarol . Or maybe you added her from Firebase Console when you created the group like all the others… but she since used the MC tool to remove herself from the group? In any case, that means that unlike me, @pinarol will have to use the MC tool to add herself to "Gravatar iOS - Prototype Build" group to be able to see build 42

### Clean up past distributions in Firebase Console to avoid confusion

To avoid confusion and inconsistencies in Firebase, I manually removed the "Gravatar SDK iOS - Prototype Build" testing group from builds 36–39 in Firebase Console, and updated the invited group for build 41 to be distributed to "Gravatar iOS - Prototype Build" instead of the "Gravatar SDK iOS - Prototype Build".

This should avoid people who might only want to test the SDK Demo in the future to unexpectedly receive an invite to the iOS app and make them confuse the Demo app with the iOS app

### Firebase Invites that were already sent

Also note that I didn't receive a _new_ "You've been invited to test GravatarApp" email from Firebase after you changed the Testing Groups. This is normal, but just wanted to mention it.

This is normal because I already accepted the invite to test this AppID / `com.gravatar.GravatarApp.prototype` bundle yesterday when I tested downloading past Prototype Build from this PR back when it was distributed with the old Testing Group. The fact that we distribute it to a new testing group now doesn't change the fact that I've already accepted the invite for this App hence not having to receive and accept an invite again even after the Testing Group change.

But that also means that anyone who were in the "GravatarSDK iOS - Prototype Build" testing group at the time your PR initially distributed the build to that group would have received an invite for testing that AppID as well back then. And based on Firebase Console they have not accepted that invite yet.

The fact that you added all of them in the "Gravatar iOS - Prototype Build" testing group in Firebase Console already will definitively help here, because that means that even if they accept that invite from the now-couple-of-days-old invite email today, they will see the builds in FAD. While if they were not yet in "Gravatar iOS - Prototype Build" (and had to go to the MC self-serve tool to add themselves to the testing group) it'd have been confusing to have received an email to test the iOS App only to not see any app in FAD if they were not in the new and only group in which that app was now distributed 😅

AliSoftware (approved):
:shipit:

mokagio:
@AliSoftware 

> I didn't have to use the MC tool to add myself to the new Testing Group, since apparently you already added all the members of the "Gravatar SDK iOS - Prototype Build" testing group to the new "Gravatar iOS - Prototype Build" testing group [...]

Correct. I should have documented it.

> That being said, when going to the Firebase Console to compare the 2 testing groups, it seems that you invited almost everyone who was part of the SDK testing group to the app testing group… except @pinarol

🤔 that's odd... I could have missed her email of course, but I checked that the two groups had the same amount of testers. See screenshot in https://github.com/Automattic/Gravatar-iOS/pull/9#discussion_r2163005808.

> To avoid confusion and inconsistencies in Firebase, I manually removed the "Gravatar SDK iOS - Prototype Build" testing group from builds 36–39 in Firebase Console,

Thanks. Always going the extra mile.

PR #10, opened 2025-06-24
App bottom tab bar

etoledom:
Closes: GRA-167

Simple PR which introduces the root TabBar for the app.

CleanShot 2025-06-24 at 15 15 05

### To test:

- Run the app
- Authenticate if not authenticate yet.
  - Check that the tab bar displays and it switches betten tabs
    - Note that Gravatar and Profile is displaying the same view for now.


GravatarApp/RootTabView.swift
```diff
@@ -0,0 +1,87 @@
+import Gravatar
+import SwiftUI
+
+struct RootTabView: View {
+    let profile: Profile
+    let onLogout: () -> Void
```
etoledom, line 6: Logouts should be handled differently. But it's here now just to connect the example logout button.
Profile might be replaced by a view model also, but so far is not needed.

GravatarApp/RootTabView.swift
```diff
@@ -0,0 +1,87 @@
...
+
+struct BackgroundColorView<Content>: View where Content: View {
+    let color: UIColor
+    let content: () -> Content
+
+    var body: some View {
+        ZStack {
+            Color(uiColor: color)
+                .ignoresSafeArea()
+            content()
```
etoledom, line 80: Full screen background color, independent of the content size.

pinarol (approved):
LGTM 👍

PR #11, opened 2025-06-24
OAuth improvements

etoledom:
This PR simplifies the OAuth logic by setting the minimum iOS version target to 17.4, and removing support for earlier versions.

Also, some improvements on development error messages.

### To test:

- Run `make setup` or `make dev`.
- Open the newly created `Secrets/Secrets.swift` file and fill it with the apps secrets.
- Run the app
- Tap on the login button
- Complete the oauth flow on web
  - Check that the logged-in screen appears with the user display name.
- Close and restart the app
  - Check that the app persisted the logged in state
- Press log-out
  - Check that you are back to the welcome screen.


pinarol (approved):
LGTM 👍

PR #12, opened 2025-06-25
Adding collapsed header view to avatar picker

etoledom:
Closes GRA-219

This PR adds a collapsed style header to the avatar picker view:

### To test:

- Run the app
- Login if needed
  - Check that the header view looks as in the designs.
  - Check that the avatar load properly.
  - Try changing the avatar at some other place and running the app again.
    - Maybe in the future we should implement pull to refresh and/or refresh on coming back from background.
  - Try logout and login with another account .


pinarol (approved):
LGTM 🎉

PR #13, opened 2025-06-25
Avatars picker grid

etoledom:
Closes GRA-168
This PR introduces the avatars grid for the avatar picker

This is heavily based on the SDK grid. Coping the structure, models, and views, but simplifying and changing a few things (hopefully for better).

- `AvatarPickerViewModel` only for AvatarPickerView.
  - On the SDK we mixed Profile and avatar picker on the same big model. Let's try to keep things apart now. 
- Modified layout and UI to match the design.
- I have implemented selecting avatar on this PR, but now it's via contextual menu instead of direct tap.

I've left some parts that are not currently used, but which we will be using soon, like Alt text editing and image upload, together with future actions on avatar tap. 

**Note:**
The main difference in layout with the SDK, is that now the grid will adapt the items size and column number to the screen width (with a maximum size), keeping always a constant horizontal padding and spacing in between elements.

I kept the design with 4 columns for the "normal" sized iPhone, and the rest will adapt in the best way it can.

Side by side comparison (never mind the top tab bar on iPad 😬 )
CleanShot 2025-06-26 at 08 47 44@2x


GravatarApp/AvatarPicker/AvatarPickerView.swift
```diff
@@ -0,0 +1,57 @@
...
+import Gravatar
+import SwiftUI
+
+struct AvatarPickerView: View {
+    let onLogout: () -> Void
+    @StateObject var avatarPickerModel: AvatarPickerViewModel
+    @State var forceRefreshHeader: Bool = false
+
+    init(avatarPickerModel: AvatarPickerViewModel, onLogout: @escaping () -> Void) {
+        self._avatarPickerModel = StateObject(wrappedValue: avatarPickerModel)
```
pinarol: Using StateObject this way is an anti-pattern which can cause hard to diagnose bugs.
@StateObject is meant to own and create its model internally. We are passing an observable model from outside so we should instead use `@ObservedObject` on avatarPickerModel.

GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -0,0 +1,376 @@
...
+            public static let half: CGFloat = 4
+            public static let single: CGFloat = 8
+            public static let split: CGFloat = 12
+            public static let double: CGFloat = 16
+            public static let medium: CGFloat = 24
+            public static let large: CGFloat = 32
+            public static let max: CGFloat = 48
+        }
+    }
+}
```
pinarol: Let's put this in its own file.

GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -0,0 +1,376 @@
...
+extension Result {
+    func error() -> Error? {
+        switch self {
+        case .failure(let error):
+            error
+        default:
+            nil
+        }
+    }
+}
```
pinarol: Let's put these in its own file

GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -0,0 +1,376 @@
...
+        if let newURL = URLComponents(string: imageURL)?.replacingQueryItem(name: "size", value: size).string {
+            return newURL
+        }
+        return imageURL
+    }
+
+    var isSelected: Bool {
+        selected ?? false
+    }
+}
```
pinarol: Let's put these in its own file.

GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -0,0 +1,376 @@
...
+            queryItems[sizeItemIndex] = newItem
+            copy.queryItems = queryItems
+        } else {
+            // Add the query item if it doesn't exist
+            copy.queryItems = (self.queryItems ?? []) + [newItem]
+        }
+
+        return copy
+    }
+}
```
pinarol: Let's put these in its own file

GravatarApp/AvatarPicker/Grid/AvatarGrid.swift
```diff
@@ -0,0 +1,99 @@
...
+            LazyVGrid(columns: columns, spacing: .avatarSpacing) {
+                ForEach(grid.avatars) { avatar in
+                    Menu {
+                        menuItems(for: avatar)
+                    } label: {
+                        avatarView(for: avatar, size: itemSize)
+                    }
+                }
+            }
+        }
```
pinarol: Instead of doing these manual calculations, let's try to make it work with GridItem's .adaptive option first. What this does it, it tries to insert as many items of the `minimum` size as possible… but it also lets them increase to the `maximum` size if needed.

So we shouldn’t have a super big size unless it’s really needed, because it’ll try to stick to the `minimum`.

GravatarApp/SharedViews/CheckmarkCircleView.swift
```diff
@@ -0,0 +1,48 @@
+import SwiftUI
+
+private enum Constants {
+    static let bluishColor = UIColor(red: 0.11, green: 0.31, blue: 0.77, alpha: 1.00)
```
pinarol: Is this color coming from designs? If so, let's define them in a central place like we had in the SDK.

GravatarApp/RootTabView.swift
```diff
@@ -83,7 +87,7 @@ struct BackgroundColorView<Content>: View where Content: View {
 }
 
 #if DEBUG // Needed when we use `Profile.testProfile on Previews`
-#Preview {
-    RootTabView(profile: .testProfile, onLogout: {})
-}
+// #Preview {
+//     RootTabView(profile: .testProfile, onLogout: {})
+// }
```
pinarol: This is commented out, did you intend to keep it this way? Or should we just remove?

pinarol (commented):
Leaving some comments but no big thing. Also there's this issue about scrolling (it doesn't scroll).


https://github.com/user-attachments/assets/d1382838-1ac8-4af1-bf8e-d65dfd14eee3

etoledom:
@pinarol - Thanks for the review.

I believe all the comments have been taken care of. Ready for another 👀

pinarol (approved):
Looks great 🎉

PR #14, opened 2025-06-26
Mobile secrets

etoledom:
Closes GRA-229

This PR setups the secrets on the CI build, in order to use OAuth.

There's a new make command `make setup-internal` meant for internal developers which will install the app secrets.
External developers can use `make setup` to generate the empty secrets file.

#### Local tests:
- Remove the local secrets file
- Run `make setup-internal`
  - Check that the secrets have been installed.
  - Run the app and test OAuth. 

#### CI tests:
- Install the app via the QR code from this PR
  - Run the app and test OAuth.


pinarol (approved):
This is great 🎉 :shipit:

PR #15, opened 2025-06-26
Collapsable header with animating content

pinarol:
Closes GRA-242 GRA-221

## Description

Implements an expandable/collapsable header that also animates its content when transitioning between these 2 states. 

You can inject any header content as soon as it implements the protocol `CollapsableHeaderViewContent`. 

Similarly, you can inject any scrollview content, it can be any SwiftUI.View or any UIKit.UIView. 

https://github.com/user-attachments/assets/1e06f7d4-1e21-42c3-98e0-0196ce821e8c

**Things to tackle further**

- Currently we set the min/max height of the header to a constant value. We need to know these 2 values to properly calculate the animation progress. However, we can try to calculate the max height during runtime since its the initial position. Min height would me more tricky but it can be also doable, although not that necessary.

- The scroll offset changes are used when calculating the animation progress and height changes. However, it can be better if we locked the scroll position during expand/collapse, this way the content doesn't hide behind the header in collapsed state. And only once the header is fully collapsed, we can let the scroll view actually scroll. Not sure how complex this would be though.

- I also wonder if the way UIViewPropertyAnimator is used in this PR can be further simplified. Currently we recreate it on some certain conditions but maybe there's a way to init it only once. 

## To Test

Go to the Profile tab and play


etoledom (approved):
Super nice! Thank you 🙏 

Let's take it from here and move forward

PR #17, opened 2025-06-27
Profile header content

etoledom:
Closes GRA-243

This PR adds the Profile content to the Profile editor header.

It also improves the header animator logic.

### To test:

- Run the app
- Go to Profile tab
- Play scrolling the table view.
  - Check that the header view behaves consistently as expected.


pinarol (approved):
Great animations 🎉 Thank you so much! 🥳

PR #18, opened 2025-06-27
Refactor WelcomeView

pinarol:
Closes part of GRA-171

## Description

There was a problem about the ownership of `avatarPickerModel`, nobody owned it. A View needs to own it and declare it as `@StateObject` to avoid lifecycle problems, the rest can use `@ObservedObject`. With this PR, `RootTabView` owns it, we can change this in the future but its suitable for the current state. 

Also the business logic inside the WelcomeView got a bit big so I extracted it out to `WelcomeViewModel`. We can reuse it even if the WelcomeView UI changes.

I replaced the ProfileService with ProfileViewModel. The old one had a @Published variable but the class didn't confirm to ObservableObject. Combine would still publish its value but to use it with SwiftUI we should confirm to ObservableObject. 

## To Test

Smoke test everything including OAuth flow


GravatarApp/Welcome/WelcomeViewModel.swift
```diff
@@ -0,0 +1,104 @@
...
+        accessToken = nil
+        await setProfile(to: nil)
+    }
+
+    func requestOAuthToken() async {
+        analytics.track(WelcomeScreenEvent.authButtonPressed)
+        oauthError = nil
+        do {
+            self.accessToken = try await oauthManager.requestAccessToken().token
+            analytics.track(WelcomeScreenEvent.authSuccess)
```
etoledom, line 104: We need to set `analytics.setUserName()` before sending this event.

GravatarApp/Welcome/WelcomeView.swift
```diff
@@ -3,31 +3,35 @@ import OAuth
...
                     }
+                    .transition(.opacity)
+                case .failure(let error):
+                    Text("Error fetching the profile: \(error)")
+                        .padding()
                 }
-                .transition(.opacity)
             } else {
                 loginView
             }
```
etoledom: Maybe something like this is a bit more readable?

```swift
        Group {
            if (viewModel.hasUser && viewModel.profileResult == nil) || viewModel.profileViewModel.isLoading == true {
                ProgressView()
            } else if let error = viewModel.profileResult?.error() {
                errorView(with: error)
            } else if
                let profile = viewModel.profileResult?.value(),
                let accessToken = viewModel.accessToken
            {
                rootView(accessToken: accessToken, profile: profile)
            } else {
                loginView
            }
        }.onAppear {
            viewModel.softLogin()
        }
```

With `rootView()` declared separatelly.
At your discretion :)

GravatarApp/Welcome/WelcomeViewModel.swift
```diff
@@ -0,0 +1,104 @@
...
+            }
+            .store(in: &cancellables)
+    }
+
+    private func setProfile(to profileResult: Result<Profile, APIError>?) async {
+        await analytics.setUserName(profileResult?.value()?.userLogin)
+
+        withAnimation {
+            self.profileResult = profileResult
+        }
```
etoledom: Since presenting `RootTabView` depends on both `profileResult` and `accessToken`, both need to be set inside `withAnimation` for the animation to run. Right now there's no animated transition.

On login is not very noticeable when the profile request is fast, but on logout is quite noticeable.

As an idea for the future: Maybe we can have a state enum with `logedOut` `loading` `loggedIn(profile, token)` `error` cases to avoid checking multiple properties per state 🤔 

I don't mind not having the animations now, but just to keep in mind 👍

GravatarApp/Welcome/WelcomeViewModel.swift
```diff
@@ -0,0 +1,104 @@
...
+            .compactMap(\.self) // ignore nil
+            .removeDuplicates()
+            .sink { [weak self] newToken in
+                guard let self else { return }
+                Task {
+                    await self.profileViewModel.fetchProfile(with: newToken)
+                }
+            }
+            .store(in: &cancellables)
+
```
etoledom: From what I see, what this does is basically:
```swift
    @Published var accessToken: String? {
        didSet {
            if let accessToken, accessToken != oldValue {
                Task {
                    await self.profileViewModel.fetchProfile(with: accessToken)
                }
            }
        }
    }
```

Any benefit from using combine here over `didSet`? I personally find `didSet` much faster to read on a glance.

The second combine setup on $accessToken (for example) seems to be beneficial to do with combine, as it puts together changes on both `accessToken` and `profileResult`, but this one I'm not that sure 🤔

etoledom (approved):
Looking great and works perfectly 🎉 

Thanks for improving the `@ObservedObject` setup.

I added a few comments, mostly small suggestion.

The one related to analytics seem important though 🙏

GravatarApp/Welcome/WelcomeView.swift
pinarol: It might be yes. This part was already there and I tried not to touch the general logic here but let me check.

GravatarApp/Welcome/WelcomeViewModel.swift
pinarol: > I personally find didSet much faster to read on a glance.

Totally understandable. 👍 I'll update.

GravatarApp/Welcome/WelcomeViewModel.swift
pinarol: Hmm let me check.

GravatarApp/Welcome/WelcomeViewModel.swift
pinarol: Having said that, since this is a `@Published` value there might be some confusions in the future about didSet vs Combine order. But for now we don't have that.

GravatarApp/Welcome/WelcomeViewModel.swift
pinarol: `WelcomeScreenEvent.authSuccess` is just OAuth success so If you ask me, it should be normal that at this stage we don't have a user name yet. Having said that, we haven't discussed these events yet so let's leave it as is for now and come back later after clarifying things.

GravatarApp/Welcome/WelcomeView.swift
pinarol: I like extracting out the rootView() but I think we should handle `profileResult ` under the same if. I find it confusing to have 2 different ifs for `viewModel.profileResult?.error()` and `viewModel.profileResult?.value()` and kind of defeats the purpose of using a single enum for profileResult`.

GravatarApp/Welcome/WelcomeViewModel.swift
pinarol: I brought back the animation here.

I tried using state enum as well but it made things more complex so I reverted.

etoledom (approved):
Looking great!

Thank you for the changes 🙏

PR #19, opened 2025-06-30
OAuth - `gravatar-global` and CODE type authentication.

etoledom:
Closes GRA-422

This PR makes two changes:

- Using the new `gravatar-global` scope.
- Using CODE type authentication, instead of TOKEN.

To test:

- Run the app.
- Logout if needed
- Sign in 
  - Check that the sign-in is successful.


pinarol (approved):
LGTM 🎉

PR #20, opened 2025-07-01
Profile editing - iter 1

pinarol:
Closes GRA-171

#Description

What's added?

- A profile editing view (mostly same with the SDK, new designs aren't ready yet)
- UserSession singleton to manage user data
- Profile header is updated when the profile is updated
- Save button is for now in the scrollable area, this will change in the upcoming PRs when things are more certain
- `.sessionExpired` notification to logout the user if the token is expired

# To test

Update your profile
Observe that the header also updates

Expire your token via https://wordpress.com/me/security/connected-applications
Try updating again
Observe that you land on the login screen


GravatarApp/ProfileEditor/EditProfileViewModel.swift
```diff
@@ -1,18 +1,68 @@
...
+            let profile = try await profileService.updateProfile(with: request, token: authToken)
+            print("success!")
+            Task { @MainActor in
+                self.userSession.updateProfile(profile)
+            }
+            // TODO: Show success toast
+        } catch APIError.responseError(let .invalidHTTPStatusCode(response, _))
+            where response.statusCode == HTTPStatus.unauthorized.rawValue
+        {
+            NotificationCenter.default.post(name: .sessionExpired, object: nil)
```
pinarol, line 52: If, in the future, it turns out that we always handle session expiration the same way we can move posting of this notification in a deeper and a single place at the networking level.

GravatarApp/ProfileEditor/EditProfileViewModel.swift
```diff
@@ -1,18 +1,64 @@
...
+            isSaving = false
+        }
+        do {
+            isSaving = true
+            let request = fields.updateRequest()
+            let profile = try await profileService.updateProfile(with: request, token: authToken)
+            Task { @MainActor in
+                self.userSession.updateProfile(profile)
+            }
+            // TODO: Show success toast
```
pinarol, line 48: In another PR, I want to bring in the toast even though we don't know how it would look like in the app yet, it's useful even when testing.

GravatarApp/ProfileEditor/EditProfileViewModel.swift
```diff
@@ -1,18 +1,64 @@
...
+        self.profileService = .init(urlSession: urlSession)
+        if let profile = userSession.profile {
+            self.fields = .init(profile: profile)
+        } else {
+            self.fields = .init()
+        }
+        userSession.$profile.sink { [weak self] profile in
+            guard let profile else { return }
+            self?.fields = .init(profile: profile)
+        }
```
etoledom: There's something I don't like about the UserSession having optional profile and token, and is the need to handle the possibility of nil values in views and models where it doesn't make sense to not have a session.

If the user doesn't have session, they should be in the welcome screen (or anywhere that it's good to not have a session). 

If the user is in the logged-in tab view (any of the tab) a profile and a token should be guaranteed.
We can achieve this by creating a UserSession on login, and passing it to the logged-in views via environment.

I made a draft PR exploring this idea here:
https://github.com/Automattic/Gravatar-iOS/pull/22

It's easy to inject a UserSession instance via `.environmentObject` to any view who needs it for previews and unit tests, and models still can get it as an init parameter. While in the running app, any view presented by the RootTabView will automatically inherit the environment. Since it's an object, it's still a shared instance, and it's Profile parameter can be updated, its update will be propagated to observers, but it will never be nullified.

etoledom: If in the future there's a view where it could be accessible without session, but with extended features having a session (i.e: future image editor), we can pass an optional UserSession? instance, instead of having its fields optionals.

etoledom (commented):
Looking good!

Just a few observations:

### Design
The implemented design looks different to the figma. If it's intended to work the design later, that's fine (specially since the figma is not definitive yet). But just noting it here.

CleanShot 2025-07-02 at 09 06 55@2x

### Save button
The save button at the bottom of the scrollable content is kinda unfriendly. Since we don't know yet where will the save button be definitively, we can have it just to have it somewhere for now. But I'd advocate to remove it from the bottom of the scrollable content at some point.

### Keyboard avoidance
We will most probably need to setup keyboard avoidance on the UIKit UIScrollView side for this one 🤔



### UserSession

I mentioned on code review a detail about this. I think it would be good to not have optional token/profile in views where a session is required. More on the [comment](https://github.com/Automattic/Gravatar-iOS/pull/20#discussion_r2179542146) itself.

GravatarApp/ProfileEditor/EditProfileViewModel.swift
etoledom: Yes, I agree it would be good to handle this in a centralized way, to avoid the need to check for this error on every single request. We can explore this later on

pinarol:
I intentionally left the design, save button and the keyboard avoidance out of this PR.

As you said we don't have the definitive designs yet. Also this PR got too big already. 

We can have UserSession with non-optional values but we can't make it singleton then, which means we would need to pass it at every step. And, passing it as an environment variable will help SwiftUI views only. I am fine trying it. I am not a big fan of using singleton anyway.

etoledom:
> I intentionally left the design, save button and the keyboard avoidance out of this PR.

Perfect 👍 

> We can have UserSession with non-optional values but we can't make it singleton then

Yes that's clear. That was done on the exploration https://github.com/Automattic/Gravatar-iOS/pull/22 . I think it's good enough. Models using UserSession will require it on init, and views can get it from the environment.

The environment is only needed to be set once at the root of the authentication-required view tree (`RootTabView` in this case).

Do you prefer to explore this on this PR? Otherwise we can merge this one and open https://github.com/Automattic/Gravatar-iOS/pull/22 as a proper PR.

pinarol:
https://github.com/Automattic/Gravatar-iOS/pull/22 is all good. I merged it into this. @etoledom

etoledom (approved):
Nice! Let's continue moving forward 🎉

PR #21, opened 2025-07-01
Avatar upload

etoledom:
Closes GRA-170

This PR adds the buttons to upload new avatars from 3 different sources:
- Camera
- Photos app
- Apple Playground (for supported devices)

Most of the types required come directly from the SDK, removing everything related to external custom image editor.

#### 1. Profile email needed for image upload (on the SDK side).

We need two breaking changes on the SDK. One of them was made here:

https://github.com/Automattic/Gravatar-SDK-iOS/pull/792

The avatar upload request (as it was implemented) required the profile's Email, but what the backend is interested in is only the email's hash. So we should use `PofileIdentifier` instead. 
As we don't have access to the email on the OAuth app, we must use the hash directly, so this change is required.

#### 2. Erasing `AvatarDetails` type from the avatar upload return.

We tried too hard to hide Avatar information. `AvatarDetails` is public, but we erase it on the return of the avatar upload, and we return `AvatarType` instead which doesn't provide enough information for the model.

~On this PR (so far) we are force-casting it, which is guaranteed to succeed, but ideally we need to get a `AvatarDetails` type from the upload.~

----
**Note**: https://github.com/Automattic/Gravatar-SDK-iOS/pull/792 has been updated to solve both problems

----

https://github.com/user-attachments/assets/b19fe7e2-53ba-4b06-801a-11ab5b0bfec4

### Test

- Run the app on a device with Apple Intelligence enabled
- Test the 3 new ways to upload images
- Test failed uploads
  - Try retry an upload


GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -231,6 +215,113 @@ class AvatarPickerViewModel: ObservableObject {
...
+
+    private func doUpload(squareImage: UIImage, localID: String, accessToken: String) async {
+        do {
+            guard let avatar = try await avatarService.upload(
+                squareImage,
+                selectionBehavior: .selectUploadedImageIfNoneSelected(for: .hashID(profile.hash)),
+                accessToken: accessToken
+            ) as? AvatarDetails else {
+                fatalError()
+            }
```
etoledom: We need to return `AvatarDetails` from `avatarService.upload()`
This is a breaking change on the SKD

pinarol: IMO, we can start doing these sorts of changes on SDK's `trunk` branch. We had a list of things for a major version anyway and we have quite some deprecations piled up. 
In the meantime if we need to fix anything we can do so in the release branch.

pinarol: We can depend point SDK dependency to a alpha/beta version or even to a commit, all better than having a fatalError() call in production.

etoledom: > all better than having a fatalError() call in production.

Yes I didn't mean to ship this fatal error to production, but I used it while we decide how to approach the SDK change. 
As mentioned in the description:
>On this PR (so far) we are force-casting it, which is guaranteed to succeed, but ideally we need to get a AvatarDetails type from the upload.

etoledom:
https://github.com/Automattic/Gravatar-SDK-iOS/pull/792 solves both problems encountered, and I've updated this PR to implement those SDK API changes

GravatarApp/AvatarPicker/AvatarUpload/ImagePickerSectionView.swift
```diff
@@ -0,0 +1,62 @@
...
+
+struct ImagePickerSectionView: View {
+    let onImageSelected: (UIImage) -> Void
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 16) {
+            VStack(alignment: .leading, spacing: 0) {
+                Text("Get a new look")
+                    .font(.headline)
+                Text("It’s been 87 days since you updated your avatar.")
```
pinarol: We don't have localization yet but I think it would still be good to define these in a way that is easy for Glotpress to pick up in the futue, which is by declaring them as NSLocalizedString. WDYT?

pinarol: > It’s been 87 days since you updated your avatar.

I know this is in the designs but we shouldn't put invalid static text I think. We don't have that info and it is unlikely we'll get it. Let's either replace it with a valid string or just remove this Text altogether until we know what to put here.

pinarol (approved):
Just small comments otherwise looks good 🎉 I don't need to check again I think.

PR #22, opened 2025-07-02
User session without optionals

etoledom:


GravatarApp/AppRoot/RootTabView.swift
```diff
@@ -4,12 +4,23 @@ import SwiftUI
...
+        self.session = session
+
+        _avatarPickerViewModel = StateObject(wrappedValue: AvatarPickerViewModel(profile: profile, authToken: accessToken))
+        _editProfileViewModel = StateObject(wrappedValue: EditProfileViewModel(userSession: session))
         self.onLogout = onLogout
+
+        _avatarPickerViewModel = StateObject(
+            wrappedValue: AvatarPickerViewModel(profile: session.profile, authToken: session.accessToken)
+        )
+        _editProfileViewModel = StateObject(wrappedValue: EditProfileViewModel(userSession: session))
```
pinarol, line 23: Seems like these models are initialized twice by mistake, but no need to change anything here, I can remove them after merging this.

pinarol:
This is pretty much the same how I would do it. I'll just merge it. Thanks!

pinarol (approved):

PR #23, opened 2025-07-02
Avatar delete action

etoledom:
Closes GRA-169

Adding the avatar deletion action

### To test:
- Run the app
- Tap on an avatar
- Select "Delete"
  - Check that the avatar has been deleted


pinarol (approved):
Looks great 🎉

PR #24, opened 2025-07-02
Add a network interceptor to handle session expired errors

pinarol:
Closes GRA-436

## Description

Adds a URLSessionProtocol implementing singleton to intercept all networking requests so we don't have to handle session expired errors separately. We might need to revisit this approach when designs are more certain but it would be easy to remove it or define exceptions for certain endpoints.

Also, I noticed that we are stuck at the loading state in case the token was expired and the app is killed. So I also tried to fixed it.

## To test

When in a logged in state
Expire your token via https://wordpress.com/me/security/connected-applications
Try any operation
Observe that you land on login

Log in
Expire your token via https://wordpress.com/me/security/connected-applications
Kill the app
Reopen the app
Observe that you land on login

### DIFF
--- a/GravatarApp/Networking/GravatarURLSession.swift
+++ b/GravatarApp/Networking/GravatarURLSession.swift
@@ -17,18 +17,20 @@ final class GravatarURLSession: URLSessionProtocol, Sendable {
     func data(for request: URLRequest) async throws -> (Data, URLResponse) {
         let result = try await urlSession.data(for: request)
         if let httpResponse = result.1 as? HTTPURLResponse,
-           httpResponse.statusCode == HTTPStatus.unauthorized.rawValue {
+           httpResponse.statusCode == HTTPStatus.unauthorized.rawValue
+        {
             Task { @MainActor in
                 NotificationCenter.default.post(name: .sessionExpired, object: nil)
             }
         }
         return result
     }
-    
+
     func upload(for request: URLRequest, from bodyData: Data) async throws -> (Data, URLResponse) {
         let result = try await urlSession.upload(for: request, from: bodyData)
         if let httpResponse = result.1 as? HTTPURLResponse,
-           httpResponse.statusCode == HTTPStatus.unauthorized.rawValue {
+           httpResponse.statusCode == HTTPStatus.unauthorized.rawValue
+        {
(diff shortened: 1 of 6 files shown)


GravatarApp/Networking/GravatarURLSession.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+import Gravatar
+
+/// Intercepts the network requests to perform common operations.
+final class GravatarURLSession: URLSessionProtocol, Sendable {
```
pinarol, line 5: I don't feel strong about the name `GravatarURLSession` but couldn't come up with something better.

GravatarApp/Welcome/WelcomeViewModel.swift
```diff
@@ -94,8 +94,10 @@ class WelcomeViewModel: ObservableObject {
     }
 
     func logout() async {
-        guard let profile = profileResult?.value() else { return }
```
pinarol, line 97: This caused a bug because the profile can't be fetched if the token is expired so the logout failed and token couldn't be removed. Updated this to use the persisted user id.

GravatarApp/Networking/GravatarURLSession.swift
```diff
@@ -0,0 +1,40 @@
...
+    static let shared = GravatarURLSession()
```
etoledom, line 6: Just curious. Any particular reason to make a shared instance instead of instantiate a new one when needed?

etoledom (approved):
Looks and works great 🎉 

Just a question out of curiosity.

:shipit:

GravatarApp/Networking/GravatarURLSession.swift
pinarol: Do you mean we could have multiple URLSessions? Because we wouldn't want that. URLSession should manage the requests and optimize them while letting them share the same cookies, cache etc.

We only want one `URLSession` instance so it's by nature a singleton. 

Or do you mean we could make *only* the `urlSession` static and create a new GravatarURLSession() everywhere?

Then it just *hides* that there's a single URLSession instance inside. Can even be confusing sometimes.

PR #25, opened 2025-07-03
Offline support

etoledom:
Closes GRA-432

This PR implements the basics of offline support:

- Persistent storage of the authenticated user profile.
- Network connectivity observer.

### Login logic:

- We search for an authenticated user's token in the keychain.
- We search for a Profile in storage
  - If there's a Profile
    - Go directly to authenticated section (tabs screen)
    - Fetch for the profile to update the stored data
    - Update the profile through UserSession which propagates the update
  - If there's no Profile
    - Show activity indicator while fetching for the Profile
    - Save the Profile in storage
    - Create the UserSession which is passed to the authenticated screens.

### Connectivity logic

- If the app starts without connectivity and a Profile is found in storage
  - Show the authenticated screens
  - The avatar grid will error since there's no connectivity to fetch them
  - The profile editor will display the latest known data
- When the connection is back:
  - Avatar list will reload automatically if an error is displayed
  - Profile fields and header will update to the last version, if there are changes coming from remote

### Future improvements

- Proper no-result display in avatar grid (possibly a "no internet connection" error message in the background)
- Block fields from being edited if there's no internet (maybe), because they will be overriden if there are changes coming from remote.
  - Alternatively, don't override the fields which have local changes.

To test:

- Test both logics described above.
- For connectivity, I recommend testing on device, as the network observer is not very reliable on the simulator.
- Test logout and login.
- Test switching accounts.

### DIFF
--- a/GravatarApp/AppRoot/GravatarAppApp.swift
+++ b/GravatarApp/AppRoot/GravatarAppApp.swift
@@ -1,11 +1,25 @@
 import OAuth
+import SwiftData
 import SwiftUI
 
 @main
 struct GravatarAppApp: App {
+    var modelContext: ModelContext
+    @State private var unrecoberableError: Error?
+
+    init() {
+        do {
+            self.modelContext = try ModelContext(ModelContainer(for: ProfileStore.self))
+        } catch {
+            print("Error creating model context: \(error)")
+            resetSwiftDataStore()
+            modelContext = try! ModelContext(ModelContainer(for: ProfileStore.self))
+        }
+    }
+
     var body: some Scene {
         WindowGroup {
-            WelcomeView()
+            WelcomeView(modelContext: modelContext)
                 .configureOAuth(
                     clientID: Secrets.clientID,
                     clientSecret: Secrets.clientSecret,
@@ -17,5 +31,21 @@ struct GravatarAppApp: App {
                     UITabBar.appearance().scrollEdgeAppearance = appearance
                 }
         }
+    }
+}
+
+private func resetSwiftDataStore() {
+    let fileManager = FileManager.default
+    let storeURL = URL.applicationSupportDirectory.appending(path: "default.store")
+
+    if fileManager.fileExists(atPath: storeURL.path) {
+        do {
+            try fileManager.removeItem(at: storeURL)
+            print("✅ SwiftData store deleted.")
+        } catch {
+            print("❌ Error deleting SwiftData store: \(error)")
+        }
+    } else {
+        print("❌ Error deleting SwiftData store: Could not find store file.")
     }
 }
--- a/GravatarApp/ProfileEditor/EditProfileViewModel.swift
+++ b/GravatarApp/ProfileEditor/EditProfileViewModel.swift
@@ -11,6 +11,8 @@ class EditProfileViewModel: ObservableObject {
     @Published var isSaving: Bool = false
     @Published var fields: ProfileFieldsModel
 
+    private var networkMonitor: any NetworkMonitor
+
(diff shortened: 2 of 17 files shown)


GravatarApp/AppRoot/GravatarAppApp.swift
```diff
@@ -1,11 +1,25 @@
...
+
+    init() {
+        do {
+            self.modelContext = try ModelContext(ModelContainer(for: ProfileStore.self))
+        } catch {
+            print("Error creating model context: \(error)")
+            resetSwiftDataStore()
+            modelContext = try! ModelContext(ModelContainer(for: ProfileStore.self))
+        }
+    }
```
etoledom, line 18: I'm not sure what to do with this `try!` here. If this fails, it's really an unrecoverable error.

Deleing the database file with `resetSwiftDataStore()`, has the same effect than re-installing the app, so this last try! should technically never fail.

etoledom: We could maybe do a last try with in-memory database, in case the error was related to writing on disk

GravatarApp/ProfileEditor/EditProfileViewModel.swift
```diff
@@ -43,11 +58,18 @@ class EditProfileViewModel: ObservableObject {
...
             // TODO: Show success toast
         } catch {
             // TODO: Show error toast
         }
     }
+
+    func fetchProfile() async {
+        do {
+            let profile = try await profileService.fetchOwnProfile(token: userSession.accessToken)
+            userSession.updateProfile(profile)
```
pinarol, line 72: Not sure about having the refresh mechanism here for the profile data. When I imagine having multiple Views that depend on the profile data, they would all be sending fetch profile requests sequentially when connection comes back. Also this code would be repeating in different ViewModels.

So I think we should decide on 1 ObservableObject as the owner of the profile data and that ObservableObject should be responsible of refreshing the profile when connection comes back. And it should be something that is created at the root level right after log in so its existence would not depend on rendering a specific screen.

And in our current structure this object seems like the `UserSession`. We can assume it's the source of truth for profile data and handle connectivity for the profile data there. 

All the other ViewModels can subscribe to profile changes from the UserSession.

Of course other ViewModels can have other data that needs refreshing so it's totally fine to have networkMonitor in them. But when we share the same data among multiple ObservableObjects we should decide which one is the source of truth I think.

etoledom: Yeah makes total sense 👍

GravatarApp/AppRoot/GravatarAppApp.swift
pinarol: I think it makes sense! I mean it will look a bit weird to nest that many error handling but it's not more important than doing the error handling! 👍

GravatarApp/ProfileEditor/EditProfileViewModel.swift
pinarol: Having said that, we don't need to block this PR. I'll approve it and leave it up to you if you want to tackle this in a separate PR or here.

pinarol (approved):
Thank you! 🎉

GravatarApp/ProfileEditor/EditProfileViewModel.swift
etoledom: Ok let's do it in a separated PR

etoledom: Created GRA-482

PR #26, opened 2025-07-03
Using `UserSession` in `AvatarPickerViewModel`

etoledom:
This PR changes `AvatarPickerViewModel` to use `UserSession` instead of Profile and AccessToken separately.

### Test:

- Smoke test avatar picker


pinarol (approved):
Looks good 👍

PR #27, opened 2025-07-04
Share view basics

etoledom:
Closes GRA-443

This PR implements the basics of the Share screen:

The header with a generated QR code, and the contact share fields.
The QR code has **no functionality yet.** This PR only focuses on layout.

I've put special effort on the layout and adaptation of the header to different screen sizes. The header will try to copy the design for iphones in portrait, but for very wide screens like iPad and landscape mode, the QR code has a maximum size.

In the case of landscape, the header height is equal to the screen, so the QR code is always fully displayed.

|   |  |
| ------------- | ------------- |
| CleanShot 2025-07-04 at 14 59 52@2x  | CleanShot 2025-07-04 at 14 56 39@2x |
| CleanShot 2025-07-04 at 14 57 01@2x  | CleanShot 2025-07-04 at 15 05 36@2x |
| CleanShot 2025-07-04 at 14 58 24@2x | CleanShot 2025-07-04 at 14 58 38@2x |
| CleanShot 2025-07-04 at 14 58 45@2x | |

### To test:

- Run the app
- Go to the Share tab
  - Check that the header display properly in any layout and orientation
  - Check that the background image updates when the user selects a different avatar.


GravatarApp/Share/ShareHeaderView.swift
```diff
@@ -0,0 +1,126 @@
...
+            .frame(height: contentHeight)
+            .clipped()
+            backgroundMaterial
+
+            VStack(spacing: 16) {
+                Text("Quickly share your public contact info by letting other scan this QR code.")
+                    .multilineTextAlignment(.center)
+                    .foregroundStyle(Color.white)
+                    .frame(width: windowWidth)
+                    .padding()
```
pinarol, line 52: This was supposed to put a padding around this text but I don't see it in the UI. Can you maybe poke a few places to see why this is happening? 🤔

pinarol (approved):
This is a good start of this page. Just 1 comment. I don't need to look again unless you want me to. Thanks 🎉

GravatarApp/Share/ShareHeaderView.swift
etoledom: I'm thinking to use `BouncyImageBackgroundHeaderView` from the [new Header PR](https://github.com/Automattic/Gravatar-iOS/pull/31) to make this header. So I'll create a PR later with this change after https://github.com/Automattic/Gravatar-iOS/pull/31 is merged. Hopefully the padding issues get solved with this change.

etoledom: GRA-483

PR #29, opened 2025-07-08
Profile editor UI refinements

pinarol:
Closes GRA-435

## Description

**What's included**

Updates the UI according to designs: Font, paddings, textfield style, colors.
Removes the contact related fields
Adds snapshot tests (also added the same SwiftUI snapshot library we use in the SDK)

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

**What's not included**

Save button changes
Keyboard avoidance

I'll wait for the SwiftUI header changes to built on.

## To Test

Check the UI
Smoke test profile editing


GravatarAppTests/Helpers/URLSessionProfileMock.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 @testable import GravatarApp
 import GravatarUI
 
-final class URLSessionAvatarPickerMock: URLSessionProtocol {
+final class URLSessionProfileMock: URLSessionProtocol {
```
etoledom: This is also extensively used by the AvatarPicker view model, so maybe a more generic name would be good.
Maybe just `URLSessionMock` is good enough?

There's also a warning here but it should be solved with a merge from trunk

GravatarAppTests/Helpers/Snapshotting+Extension.swift
```diff
@@ -0,0 +1,18 @@
...
+import SwiftUI
+import XCTest
+
+extension Snapshotting where Value: SwiftUI.View, Format == UIImage {
+    static func testStrategy(userInterfaceStyle: UIUserInterfaceStyle = .light, layout: SwiftUISnapshotLayout = .sizeThatFits) -> Self {
+        let deviceConfig: ViewImageConfig = .iPhone13
+        let traits = UITraitCollection(traitsFrom: [
+            UITraitCollection(displayScale: deviceConfig.traits.displayScale),
+            UITraitCollection(userInterfaceStyle: userInterfaceStyle),
+        ])
```
etoledom: I see a warning here also:
`'init(traitsFrom:)' was deprecated in iOS 17.0: Use UITraitCollection.init(mutations:) and UITraitCollection.modifyingTraits(_:) to create and modify trait collections`

etoledom (approved):
Looking great!

Just a couple minor observations. No need to check again.

Thank you for incorporating the Snapshot Tests! 🙏

GravatarAppTests/Helpers/URLSessionProfileMock.swift
pinarol: I thought URLSessionProfileMock is generic and covers avatar related things because avatar is part of profile too but apparently it doesn't come across as that generic. I'll update. 👍

GravatarAppTests/Helpers/Snapshotting+Extension.swift
pinarol: Fixed 👍

GravatarAppTests/Helpers/URLSessionProfileMock.swift
etoledom: Yeah, since we have an specific "Profile" section separated from "AvatarPicker" section, the first thought that came to my mind was "Profile-section-specific URL Session". It didn't occur to me to think of Profile as global in this context.

Thank you for the update 🙏

PR #30, opened 2025-07-08
Add toast component

pinarol:
Closes GRA-479

##Description

Brings the toast component from SDK
Adjusts the colors
Removes the `.warning` type
Adds snapshot tests

`.info` and `.error` toasts have the same look in the new desings but I wanted to keep these cases anyway. It would be easy to update the UI in the future if we go for a UI adjustment.

##

Check the snapshots/previews
CI green


GravatarApp/Extensions/View+Extenstions.swift
```diff
@@ -0,0 +1,12 @@
...
+extension View {
+    @ViewBuilder
+    func `if`(_ condition: Bool, transform: (Self) -> some View) -> some View {
+        if condition {
+            transform(self)
+        } else {
+            self
+        }
+    }
+}
```
etoledom: Hah I also added this one (a couple of times on different PRs)

Note: `trunk` has a file `View+extensions.swift`. When merging back let's be careful to not repeat files.
Also, let's agree on either `+extensions` or `+Extensions`. I don't really care which one :) 

Maybe uppercase E is more common actually 🤔

etoledom (approved):
Looks great 🎉 

Just one small thing to consider when updating from trunk.
:shipit:

GravatarApp/Extensions/View+Extenstions.swift
pinarol: Yeah I think uppercase is more common.

PR #31, opened 2025-07-08
New animated header

etoledom:
Closes GRA-480
Closes GRA-434

This PR introduces a simplified header animation purely written in SwiftUI, which replaces the previous `UIViewPropertyAnimator` animated header.

Both `AvatarPicker` and `ProfileEditor` headers have been implemented.

https://github.com/user-attachments/assets/c0a97176-b6d2-4215-a33b-4099b4cc4d75

The main classes are:
### `BouncyImageBackgroundHeaderView`
Stand alone "background" image view with the bouncy effect.
This one can be later on used in the Share section header.

### `AnimatedHeaderScrollView`
Which uses `BouncyImageBackgroundHeaderView` as the scrollable table header view, and animates in a sticky header view when the table header view scrolls up out of view.

## Features implemented:

- Animation behaviors: 
  - Automatic (The sticky header animation runs automatically)
  - Interactive ( The sticky header animation runs alongside the offset of the scroll view)
- System-based keyboard avoidance
- Dynamic types and dynamic height
- Landscape and iPad screens
- Modularization to improve re-usability
- Previews on all the new views to facilitate future changes and to understand how to use them.

Since all headers are based in the same layout system of the  `AnimatedHeaderScrollView`, it's easy to get a nice transition effect between avatar picker and profile view, where it looks like the main avatar image is the same, when both header are either expanded or collapsed, in both landscape and portrait.

## What's not considered:
There are some details that I see could improve, but this PR is big enough already:
- Lodaing state of the headers avatars
- Action buttons on the Profile Editor
- Animated transition from loading to loaded avatar in the headers
- A special case on the iPhone SE where the sticky header content seem to be a bit too close to the top

CleanShot 2025-07-08 at 20 51 16@2x

## To test:

- Try everything you can to break the headers.


GravatarApp/CommonViews/AnimatedHeaderScrollView/AnimatedHeaderScrollView.swift
```diff
@@ -0,0 +1,161 @@
+import SwiftUI
+
+struct AnimatedHeaderScrollView<ContentView, ScrollableHeader, StickyHeader, MenuItems>: View
+    where ContentView: View, ScrollableHeader: View, StickyHeader: View, MenuItems: View
```
pinarol: Nit: we can just do this without the `where` clause.

```
struct AnimatedHeaderScrollView<ContentView: View, ScrollableHeader: View, StickyHeader: View, MenuItems: View>: View
```

GravatarApp/AvatarPicker/AvatarPickerView.swift
```diff
@@ -9,31 +9,58 @@ struct AvatarPickerView: View {
     @State private var forceRefreshHeader: Bool = false
     @State private var avatarToDelete: AvatarImageModel?
 
+    @State private var theText: String = ""
```
pinarol: Not used from anywhere thus let's remove.

pinarol (approved):
Looking great 🎉 🙇

GravatarApp/AvatarPicker/AvatarPickerView.swift
etoledom: Ups! Debugging leftovers 😅

PR #32, opened 2025-07-09
Make `UserSession` listen to connectivity changes

etoledom:
Closes GRA-482

On this PR: 

UserSession will listen for connectivity changes, fetch for an updated profile when connectivity is back from disconnected, and update the profile if changes are found.

To test:

- Run the app on **device** (important since connectivity changes are unreliable on simulator)
- Go to the profile editor section.
- Turn off internet connection from the device
- Update your display name (or any field) on web
- Connect the device back to internet.
  - Check that the profile updates automatically with the latest data.


pinarol (approved):
Looks good 🎉

PR #33, opened 2025-07-09
Add toast component

pinarol:
Closes GRA-479

##Description

Brings the toast component from SDK
Adjusts the colors
Removes the `.warning` type
Adds snapshot tests

`.info` and `.error` toasts have the same look in the new designs but I wanted to keep these cases anyway. It would be easy to update the UI in the future if we go for a UI adjustment.

##

Check the snapshots/previews
CI green


pinarol:
Opening the same PR with https://github.com/Automattic/Gravatar-iOS/pull/30 since that one was not merged into `trunk`.

pinarol:
I'll just merge this because it's a copy of the [approved PR.](https://github.com/Automattic/Gravatar-iOS/pull/30)

PR #36, opened 2025-07-10
[Avatar Picker] Implementing avatar share action

etoledom:
Closes GRA-489

This PR implements the avatar share action, in a similar way to the SDK.

**Note**: Sadly `ShareLink` won't work for us. It expects to know the image URL in advance to render the button which presents the share sheet, but in our case, the button tap will trigger a request which finally will give us the URL.

Since it's the first time we implement an usage of the toast, I've also added the toast container to the Avatar Picker.

I've also added `NSPhotoLibraryAddUsageDescription` to have the Save to Photos option.

### To test:

- Share an avatar
- Try without internet to see the toast.


GravatarApp.xcodeproj/project.pbxproj
```diff
@@ -369,6 +369,7 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				INCLUDED_SOURCE_FILE_NAMES = "";
 				INFOPLIST_KEY_NSCameraUsageDescription = "Upload new avatars";
+				INFOPLIST_KEY_NSPhotoLibraryAddUsageDescription = "Save avatars to your Photos app";
```
pinarol: Users may not be that familiar with the "app" word. Some alternatives I could think of:

Save avatars to your Photos
Save to Photos
Save avatars to your photo library

GravatarApp/Extensions/View+Extensions.swift
```diff
@@ -48,4 +48,12 @@ extension View {
             .padding(.DS.Padding.split)
             .borders(colorScheme: colorScheme)
     }
+
+    func addToastContainer(manager: ToastManager) -> some View {
```
pinarol, line 52: This modifier worries me a bit since it wraps the view with a ZStack and you could easily miss that it does that. It's like a hidden bomb.
Can we put the ZStack explicitly to where necessary and remove this modifier?

GravatarApp/Extensions/View+Extensions.swift
```diff
@@ -48,4 +48,12 @@ extension View {
...
+        ZStack {
+            self
+            ToastContainerView(toastManager: manager)
+                .padding(.horizontal, 16)
```
pinarol: nit: let's use .DS.Padding... instead of 16

GravatarApp/AvatarPicker/AvatarPickerView.swift
```diff
@@ -62,11 +63,13 @@ struct AvatarPickerView: View {
                 }
             }
         }
+        .addToastContainer(manager: avatarPickerModel.toastManager)
```
pinarol: I have second thoughts about where to put the toast container. 🤔

I think we should put it in RootTabView but in the content area so it appears above the tab bar. And it would have this continuity when you change the tabs. Jetpack does the same thing. In that case ToastManager would be owned by RootTabView, be tagged with a `@StateObject` and passed to other view models from there.

GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -11,6 +11,8 @@ class AvatarPickerViewModel: ObservableObject {
 
     private var avatarSelectionTask: Task<AvatarDetails?, Never>?
 
+    @ObservedObject var toastManager: ToastManager = .init()
```
pinarol: This `@ObservedObject` here doesn't do anything meaningful in a view model I think and can be removed, as well as from the SDK but we can do it later.

GravatarApp.xcodeproj/project.pbxproj
etoledom: Let's go with 
> Save avatars to your photo library

GravatarApp/AvatarPicker/AvatarPickerView.swift
etoledom: Let's try this out 👍

pinarol (approved):
Looks great 🎉

PR #37, opened 2025-07-11
Pull to refresh

etoledom:
Closes GRA-494
Closes GRA-490

This PR implements a pull to refresh mechanism on the bouncy header scroll view. And implements it in both Avatar Picker and Profile Editor.

What this PR does not consider is the override of unsaved changes on the profile editor.

https://github.com/user-attachments/assets/c913853f-73de-4387-bac3-5272fa5ff1ea

### To test:

- Run the app
- Pull to refresh :D  
- To consider:
  - Length needed to pull to activate the refresh 
  - Fluidity of animations
  - No unexpected feelings


GravatarApp/CommonViews/AnimatedHeaderScrollView/BouncyImageBackgroundHeaderView.swift
```diff
@@ -2,7 +2,7 @@ import GravatarUI
 import SwiftUI
 
 struct BouncyImageBackgroundHeaderView<Content>: View where Content: View {
-    let topSafeArea: CGFloat
+    let topPadding: CGFloat
```
etoledom, line 5: Bouncy will now receive the complete top padding, independent of if it's top safe are or something else.

This makes it more customizable, like incrementing the view height in special circumstances (like adding space for the refresh control while loading)

Is up to its parent to calculate correctly Bouncy's top padding acording to existence or not of top safe areas.

GravatarApp/CommonViews/AnimatedHeaderScrollView/RefreshControl.swift
```diff
@@ -0,0 +1,34 @@
...
+    func makeUIView(context: Context) -> UIActivityIndicatorView {
+        let refresh = UIActivityIndicatorView(style: .large)
+        if animated {
+            refresh.startAnimating()
+        } else {
+            refresh.stopAnimating()
+        }
+        refresh.hidesWhenStopped = false
+        refresh.alpha = opacity
+        refresh.color = .systemBackground
```
pinarol, line 16: Spinner's color in dark mode appears black and it looks a bit incompatible with other labels since they are all white. I think we should match it with other foreground colors in the header, which is white? 
(Also, semantically it doesn't match with systemBackground as a general practice since it needs to be noticeable on a background.)

Simulator Screenshot - iPhone 16 - 2025-07-14 at 13 50 44

pinarol (approved):
This looks and works amazing! 🎉 

I just have 1 comment about the spinner color in dark mode but I don't need to look again.

PR #38, opened 2025-07-11
Profile editing - input field states

pinarol:
Closes GRA-495

## Description

Implements UI states for focused and unsaved changes for the input fields in profile editing. Also added interactive keyboard dismissal.

Simulator Screenshot - iPhone 16 - 2025-07-11 at 18 32 36

## To test

Go to profile editing:
If the field has changes, it should have a blue-ish background
If the field is focused, it should have a blue border
If any field has changes, the save button should be activated

Extra check: if the field is optional like firstName and lastName, we should make sure their empty states aren't interpreted as unsaved changes. (These should be covered by unit tests but let's also check manually.)


GravatarApp/ProfileEditor/ProfileEditContentView.swift
```diff
@@ -117,13 +119,13 @@ struct ProfileEditContentView: View {
     }
 
     private func inputField(
-        for title: String,
+        _ field: ProfileField,
         footerText: AttributedString? = nil,
         value: Binding<String>,
         isLarge: Bool = false
     ) -> some View {
```
etoledom, line 126: Just for awareness:

We will reuse these fields (at least the style) in the share screen and in the alt text editor.
Not needed on this PR, but we will eventually need to extract this component as a Common View, together with all related constants currently defined on the `ProfileEditContentView` level.

GravatarApp/ProfileEditor/ProfileEditContentView.swift
```diff
@@ -147,9 +151,11 @@ struct ProfileEditContentView: View {
                 .font(Constants.textInputFont)
                 .padding(.DS.Padding.split)
                 .disabled(viewModel.isSaving)
-                .accessibilityLabel(title)
-                .background(Color(uiColor: Constants.textBackgroundColor))
+                .accessibilityLabel(field.localizedTitle)
+                .backgroundColor(field, value: value.wrappedValue, viewModel: viewModel)
                 .cornerRadius(Constants.textInputCornerRadius)
+                .focused($focusedField, equals: field)
+                .focusedBorder(focusedField: $focusedField, field: field)
```
etoledom: I thought about adding a very short animation for the border color, but when I tested, even though it does look nicer, it gave me a kinda draggy sensation, even with an almost not noticeable at 0.1s, which was the idea.

I think is better to keep this one without animation (so, as it is now 👍 )

etoledom (approved):
Working great!

Just one question about the design implementation:

The design shows a transparent background for the focused field:

CleanShot 2025-07-11 at 21 34 54@2x


Was it intended to remove this?

GravatarApp/ProfileEditor/ProfileEditContentView.swift
```diff
@@ -165,13 +171,29 @@ struct ProfileEditContentView: View {
...
-            borderWidth: 1
+            RoundedRectangle(cornerRadius: ProfileEditContentView.Constants.textInputCornerRadius),
+            borderColor: Color(uiColor: ProfileEditContentView.Constants.focusedTextBorderColor),
+            borderWidth: focusedField.wrappedValue == field ? 2 : 0
         )
     }
+
+    fileprivate func backgroundColor(
+        _ field: ProfileField,
+        value: String,
```
etoledom: `value` here is not being used

GravatarApp/ProfileEditor/ProfileEditContentView.swift
pinarol: I extracted out the text input component as `StatefulTextField` and it got much easier to do snapshot tests for its different states. It'll be easy to reuse it in the QR page too.
 
It doesn't look like we'll reuse the title and the footer for now but if it's needed then we can extract another component I think.

pinarol:
> The design shows a transparent background for the focused field:

Oh I missed that! I now updated it. 
Although, the transparent background lasts until you type so I don't know how I feel about it in practice.

GravatarApp/ProfileEditor/ProfileEditContentView.swift
pinarol: Removed

GravatarApp/ProfileEditor/ProfileEditContentView.swift
etoledom: Amazing! Thank you 🙏

etoledom:
> Although, the transparent background lasts until you type so I don't know how I feel about it in practice.

Yes I was thinking about that. Maybe it's too many changes of color.
We can wait for a design review and update this later on after v1

PR #40, opened 2025-07-14
[Profile Editor] Save profile buttons

etoledom:
Closes GRA-493

This PR implements the save profile button from the designs, and removes the old one.

https://github.com/user-attachments/assets/4fffef18-edf5-4955-a511-cff4d5126312

To test:

- Go to the profile editor
- Make a change to any of the fields
- Test both Save and Cancel buttons


pinarol (approved):
Looks great 🎉

PR #41, opened 2025-07-15
Add avatar loading error view & empty avatars label

pinarol:
Closes GRA-519 GRA-492

## Description
2 changes:

Shows the avatar fetching error if the grid is empty:

Simulator Screenshot - iPhone 16 - 2025-07-15 at 15 11 15

Shows a different subtext if the avatar fetching call is successful but the result is just empty:

Simulator Screenshot - iPhone 16 - 2025-07-15 at 15 14 00

## Test

Test 1

Disconnect from internet and open the app
Observe the new error in place of the avatars grid
Reconnect and tap Retry
Observe avatars are loading
Disconnect again and pull to refresh
Observe that even though the error toast appears the avatar grid is still there

Test 2

Delete all the avatars
Observe the new "Your avatars will show up here" subtext


GravatarApp/DesignSystem.swift
```diff
@@ -28,6 +34,24 @@ struct ActionButtonStyle: ButtonStyle {
     enum Style {
         case primary
         case secondary
+
+        var backgroundColor: Color {
+            switch self {
+            case .primary:
+                Color(uiColor: UIColor(light: .black, dark: .DS.almostWhite))
```
pinarol: Trying this almost white color for the dark mode. I can revert if this doesn't look good to you. @etoledom  It's a little better than full white to my eyes. And it looks more in harmony with the toast.

etoledom: Looks great! We could call it  `contrastingBackgroundColor` or `oppositeBackgroundColor`, to differentiate it from the normal background color.

I mean, as a global DS constant to be reused in the toast. Same with the foreground color probably 🤔

etoledom (approved):
Working great! 🎉 

Just a small suggestion on code review. All different states worked perfectly on my tests 💪 

:shipit:

GravatarApp/DesignSystem.swift
pinarol: I like it! 👍 `oppositeBackgroundColor`

PR #42, opened 2025-07-15
Welcome screen design

etoledom:
Closes GRA-508

This PR implements the design for the Welcome screen, together with the error state.

CleanShot 2025-07-15 at 20 25 37@2x
CleanShot 2025-07-15 at 20 25 57@2x
CleanShot 2025-07-15 at 20 26 14@2x
CleanShot 2025-07-15 at 20 30 03@2x

### To test:

- Check that the design corresponds to Figma
- Deny access on OAuth
  - Check that `You need to log in to Gravatar.com` message is displayed.
    - This message mimics the access_denied error message from OAuth which says `You need to log in to WordPress.com`.
- Run Proxyman (or similar proxy) and try the OAuth process.
  - Check that you get a SSL related error.
    - I thought this is important since it could happen to users if they connect to insecure networks.
- Test the Profile error (this one is more tricky)
  - Replace `WelcomeViewModel.swift:46` with ` throw APIError.responseError(reason: .unexpected(NSError(domain: "", code: 1)))`
  - This forces to present the Profile error UI, still with the profile loading time.
  - Try to Log in
    - Check that the Profile error UI appears.
    - Check that `Retry` button tries to load the Profile again.
    - Check that `Try with another account` button moves the UI to a clean state and triggers the OAuth flow.


GravatarApp/Welcome/WelcomeView.swift
```diff
@@ -5,24 +5,20 @@ import SwiftUI
 
 struct WelcomeView: View {
     @StateObject private var viewModel: WelcomeViewModel
-    private var modelContext: ModelContext
 
-    init(modelContext: ModelContext) {
-        self.modelContext = modelContext
-        _viewModel = .init(wrappedValue: WelcomeViewModel(context: modelContext))
+    init(viewModel: WelcomeViewModel, userDefaults: UserDefaults = .standard) {
+        _viewModel = .init(wrappedValue: viewModel)
```
pinarol, line 10: Can we update this as:

```
_viewModel = .init(wrappedValue: WelcomeViewModel(context, userDefaults))
```

It won't work differently but it will make it clear who owns the model. Usually, if a View takes an observable object in the initializer the general expectation would be that it doesn't own it and uses it as @ObservedObject, unlike our case here. Here, WelcomeView owns the object. 

If we use more of this throughout the codebase, it can cause confusions around the ownership responsibilities and there’s a chance this habit could lead to tricky bugs down the road.

pinarol (commented):
Looks pretty good overall. 
I am just seeing this buggy state with steps:

- login
- profile fetching fails
- retry 
- profile fetching succeeds
- Logout
Observe: The login screen still has the error state
Expected: The login screen appears in clean state

(I used proxyman's breakpoints to simulate a failing profile fetching call)

https://github.com/user-attachments/assets/b2c01ded-eedf-4b00-b468-9c70e9577b22

GravatarApp/Welcome/WelcomeView.swift
etoledom: It was necessary to manipulate the view model in the unit tests. If the view model is created internally, we don't have access to it to test different states.

pinarol: In that case we can make this one `@ObservableObject` and have the parent own it instead I think. https://github.com/Automattic/Gravatar-iOS/compare/etoledom/welcome-screen-design...wppinar/testWelcomeViewModel?expand=1

pinarol (commented):
I can still produce this issue https://github.com/Automattic/Gravatar-iOS/pull/42#pullrequestreview-3025216470

Plus, these animations look a bit off to me. The button's background color moves differently than its label and looks like a bug. Can we maybe not have any animation for these transitions?

https://github.com/user-attachments/assets/81d1f4ed-3042-4776-9b8c-c3ed1eeb734d




https://github.com/user-attachments/assets/ae64d908-0608-4239-bfea-1ce5ca2607a8

etoledom:
> I can still produce this issue https://github.com/Automattic/Gravatar-iOS/pull/42#pullrequestreview-3025216470

I made a fix for this, then it got removed probably while cleaning up before pushing 🤦 

I've also made some modifications to the animation. Not sure if there's a better way, but now all elements seem to move together.

@pinarol - ready for another look

pinarol (approved):
The issue is resolved and the animations looks much better 🎉 Thank you! :shipit:

PR #44, opened 2025-07-16
Open profile url from Profile editor header

etoledom:
Closes GRA-520

Opens the profile url in an external browser. The browser opened will be the default predefined by the user.

### To test:

- Go to the profile editor
- Tap on the header link
  - Check that the profile page is open in an external browser


GravatarApp/ProfileEditor/Header/ProfileEditorScrollableHeaderView.swift
```diff
@@ -7,6 +7,8 @@ struct ProfileEditorScrollableHeaderView: View {
     let imageURL: URL?
     @Binding var forceRefresh: Bool
 
+    @Environment(\.openURL) var openURL
```
pinarol, line 10: TIL!

pinarol (approved):
Looks good 🥳

PR #45, opened 2025-07-16
Main menu

etoledom:
Closes GRA-543

This PR implements the Main Menu with all the actions.

It also implements a system to present views modally, in a similar way than `UIViewController - over context presentation` with a custom animation.

This is the only way I found on pure swift to achieve this effect. Other possibility is to use UIKit presentations directly, which seemed more complicated.

**NOTE:** Not considered on this PR are the real links for Terms of Services and Privacy Policy.

CleanShot 2025-07-16 at 20 59 56@2x
CleanShot 2025-07-16 at 21 00 06@2x

### To test:

- Run the app
- Tap on the menu button
- Test all the options


GravatarApp/CommonViews/ModalPresentation/ModalView.swift
```diff
@@ -0,0 +1,42 @@
...
+        }
+        .scrollBounceBehavior(.basedOnSize)
+        .background(.regularMaterial)
+        // Render solid color background only on snapshot tests.
+        // Background materials are not rended.
+        .if(ProcessInfo.processInfo.isSnapshotTesting) { view in
+            view.background(
+                Color(uiColor: .secondarySystemBackground)
+            )
+        }
```
pinarol: It's not too bad but for this specific case I would suggest an environment variable like this:

https://github.com/Automattic/Gravatar-iOS/compare/etoledom/main-menu...wppinar/modal-background-environment-var?expand=1

WDYT?

(One snapshot changes because it was applying the .background(...) after another .background(...) and now there's only 1 .background(...) I think)

etoledom: Oh that's much cleaner. I like it!
Thank you 🙏

etoledom:
I've merged the env variable branch into this one and did a bit of clean up.
Thank you for the suggestion @pinarol  🙏

pinarol (approved):
Looks good 🎉

PR #46, opened 2025-07-16
Fixes label alignment in the condensed profile header in case data is missing

pinarol:
Closes GRA-545

## Description

Fixes this issue where the label is not centered to the avatar if data is missing:
<image src="https://github.com/user-attachments/assets/e0d22633-8ac8-49b2-b56e-90eaa75f2210" width=300/>

And while on it I added some snapshot tests.

## Test

Remove the necessary profile data to check this ☝️


etoledom (approved):
Thank you 🙏 

The reason I decided to go with alignment top was this one:

CleanShot 2025-07-17 at 11 10 00@2x

When the font grows too much, it looks better if the avatar stays at the top.
But this is much less common and it has other solutions (and it's not urgent).

So let's go with this one 👍 

:shipit:

pinarol:
> The reason I decided to go with alignment top was this one:

Ah I see. 

> But this is much less common and it has other solutions (and it's not urgent).

Yeah I agree. I think the lesser bad here is to just center it.

PR #48, opened 2025-07-17
Add readable content width constraint

pinarol:
Closes GRA-553

## Description

Adds the readable content width constraint https://developer.apple.com/documentation/uikit/uiview/readablecontentguide 

Before:
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-17 at 12 43 13

After:

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-17 at 12 40 31

## To test

Check the profile and share tabs to see the constraint applies
Check with different kinds of devices to see if there's anything odd looking


etoledom (approved):
Looking great!

Thank you 🙏

PR #49, opened 2025-07-17
Add app icon

pinarol:
Closes GRA-558

## Description

Adds the app icon. 1024x1024 PNG.

## Test

Check the app icon in dark/light mode, it should be the same.


etoledom (approved):
So nice! Thank you 🙏

PR #50, opened 2025-07-17
Share scree header design

etoledom:
Closes GRA-483

This PR implements the latest design of the Share header.

Only the UI has been implemented without any functionality (appart from the Main Menu).

**Note:** There's a small main thread blockage while the QR code is generated. This is meant to be fixed when we dynamically generate the real QR code.

CleanShot 2025-07-17 at 16 58 52@2x

### To test:

- Check that the share screen header looks the same than the designs


etoledom:
Note: All the updated snapshot tests are because I've updated the size of the main menu button to be 44pts as in the design.

pinarol (approved):
Looks good 🎉 Thanks! 🙇

PR #51, opened 2025-07-17
Avoid welcome screen flash on soft login

etoledom:
Closes GRA-560

After https://github.com/Automattic/Gravatar-iOS/pull/42, I noticed that the Welcome screen was flashing for a split second during the soft login (looking for the Profile in database).

This PR avoids this flash by removing animations when the Profile is found locally, and going directly to inside the app.

### To test:

- Run the app being logged in.
  - Check that there's no initial flashing of the welcome screen


pinarol (approved):
Looks good 🥳

PR #52, opened 2025-07-18
Share view content

etoledom:
Closes GRA-555

This PR implements the content of the Share view as in the design.

I've excluded the "Domain" and "X account" since I needed more information on how to approach them.
It's not determined either how to display a fields which contains no data.

**Functionality implemented:**
- Filling up real data.
- Storing selections
- Storing private data locally (email, phone number)
- Contact vCard preview (fake data)

https://github.com/user-attachments/assets/e09a7df5-1d64-47a0-96dc-740a1e3fc51a

### To test

This PR is mostly about the UI.
- Check that the implemented UI corresponds to the design.


GravatarApp/Share/ShareContentView.swift
```diff
@@ -1,97 +1,172 @@
...
 import SwiftUI
 
 struct ShareContentView: View {
     @ObservedObject var viewModel: ShareViewModel
-    @EnvironmentObject var userSession: UserSession
 
-    @State var forceRefresh: Bool = false
-    @State var toggleOn: Bool = false
+    var body: some View {
+        VStack(spacing: 16) {
```
pinarol: nit: .DS.Padding.double

GravatarApp/Share/ShareView.swift
```diff
@@ -0,0 +1,46 @@
...
+            with: .hashID(viewModel.userSession.profile.hash),
+            options: .init(preferredSize: .pixels(256))
+        )?.url
+    }
+
+    var body: some View {
+        GeometryReader { geometry in
+            ScrollView {
+                ShareHeaderView(
+                    profile: viewModel.userSession.profile,
```
pinarol: SwiftUI won't observe nested ObservableObject changes automatically so we need to do something like this:

https://github.com/Automattic/Gravatar-iOS/pull/new/wppinar/receive-updates-from-usersession

pinarol (approved):
Looks good, just one change will be needed to propagate the profile updates to the QR tab and I already made a quick PR for that. I don't need to check again I think. 👍

GravatarApp/Share/ShareView.swift
pinarol: A better link https://github.com/Automattic/Gravatar-iOS/compare/etoledom/share-content...wppinar/receive-updates-from-usersession?expand=1

pinarol: A draft PR https://github.com/Automattic/Gravatar-iOS/pull/53

etoledom: Thank you! Merged 🙏 

I missed this. Probably I didn't even tried updates since my focus was on the UI. This change works perfectly 👍

PR #54, opened 2025-07-21
Share vcard QR generator

etoledom:
Closes GRA-445

On this PR:
- Generating QR code with the profile data selected by the user to be shared.
- Added some missing fields
  - The Verified Accounts fields will be added in a future PR
- ShareViewModel now owns the QR generator
- ShareHeaderView takes the QR image as a parameter (not generating it internally anymore)
- The QR Code is generated asynchronically to avoid blocking the main thread. 
- The QR code is refreshed automatically every time the user flips a switch, the email or phone number fields lose focus (after an edition), or the profile changes.
- Empty share field state
  - The corresponding switch state will be handled in a future PR
  - Hiding the field if empty / adding link button to the Profile editor will be handled in a future PR 

https://github.com/user-attachments/assets/c9dfba0b-1f59-48b9-b5c6-9ce5d4e9c0fd

### To test:

- Check via Preview or scanning the QR code that the shared data corresponds with the selected data by the user
- Check that the selected data persists over app restarts
- Check that the QR code refreshes with the lates data in every possible case


GravatarApp/Share/QRCode/VCardGenerator.swift
```diff
@@ -0,0 +1,39 @@
...
+    let displayName: String
+    let organization: String
+    let jobTitle: String
+    let phoneNumber: String
+    let email: String
+    let profileURL: String
+}
+
+struct VCardGenerator {
+    let profile: Profile
```
pinarol: profile isn't used, can be removed?

GravatarApp/Share/QRCode/VCardGenerator.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    func generate(with model: VCardModel) -> String {
```
pinarol: Could you add a unit test for this one?

pinarol: Also, it seems like we don't need a separate struct like VCardGenerator imo, no state needs to be stored here and the functionality could go to VCardModel? These extra types aren't always making things easy to understand.

GravatarApp/Share/QRCode/QRGenerator.swift
```diff
@@ -1,33 +1,44 @@
...
-    private let profile: Profile
-
-    private let context = CIContext()
-    private let filter = CIFilter.qrCodeGenerator()
-
-    init(profile: Profile) {
-        self.profile = profile
+import SwiftUI
+
+@Observable
```
pinarol: No stored variables here so we don't need it to be `@Observable` ?

GravatarApp/Share/QRCode/QRGenerator.swift
```diff
@@ -1,33 +1,44 @@
...
     }
 
-    var contactQRCode: UIImage {
-        generateQRCode(from: "contact:\(profile.displayName)")
+    static var fallbakImage: some View {
+        Image(systemName: "qrcode")
+            .resizable()
+            .foregroundStyle(.black)
+            .padding()
+            .background(Color.white)
```
pinarol: IMO we don't need a separate QRGenerator class for these. 

The `func generateQRCode(from string: String) async -> Image` could be another utility under `QRCodeRenderer`. 

And `fallbakImage` could just be inside ShareView since this sort of style makes sense on that screen anyway. 

Having these additional types without a clear separation of concerns makes things a bit difficult to understand imo. It took me a while to get the relationship between these.

GravatarApp/Share/ShareFields/ShareField.swift
```diff
@@ -14,7 +14,14 @@ struct ShareField: View {
...
                     .font(.caption)
                     .foregroundStyle(.secondary)
-                Text(value)
+                if value.isEmpty {
+                    Text("- \(Localized.noData) -")
+                        .font(.subheadline)
+                        .italic()
+                        .foregroundStyle(.secondary)
+                } else {
+                    Text(value)
```
pinarol, line 23: I remember like we would remove the row if there's no data. 🤔

pinarol: The active/inactive states of the empty rows also don't match and looks like a bug.

GravatarApp/Share/QRCode/QRGenerator.swift
etoledom: > And fallbakImage could just be inside ShareView since this sort of style makes sense on that screen anyway.

I'm using this extensively in previews and unit tests. Not just in `ShareView`, but also in the `ShareHeaderView`.
Makes sense to me to keep it constrained to the QR-related code, though I'd agree that `QRGenerator` shouldn't care about UI styles 🤔 

I'm thinking we could move this to a static var on `ShareHeaderView` or just `Image`. I don't particularly like these options either though.

GravatarApp/Share/QRCode/VCardGenerator.swift
etoledom: > Could you add a unit test for this one?

This has some changes in upcoming PRs, so I'll add them on the latest PR directly

GravatarApp/Share/QRCode/QRGenerator.swift
etoledom: > I'm using this extensively in previews and unit tests.

This is more noticeable in upcoming PRs

etoledom:
@pinarol - Ready for another look 👀

pinarol (approved):
let's :shipit:  🎉

PR #55, opened 2025-07-22
Adding expand QR code button action

etoledom:
Closes GRA-556

This PR implements the expand QR button on the share screen:

https://github.com/user-attachments/assets/9a9e9542-3de4-431e-bad9-55fbf882e058

### To test:

- Check that the expand button works as expected
- On iPad, a maximum QR image size was chosen. Check that the size feels right
- Testing on device:
  - The brightness will increase to maximum when showing the QR full screen, and it should come back to the previous one when the full screen view is closed.


GravatarApp/Share/ShareHeaderView.swift
```diff
@@ -60,6 +61,16 @@ struct ShareHeaderView<QRImage: View>: View {
             .padding(.horizontal, .Global.contentHorizontalPadding)
         }
         .contentWidthtReader($windowWidth)
+        .sheet(isPresented: $presentFullScreen) {
```
pinarol: This looks like this in iPad:

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-22 at 15 16 03

And it's not being very helpful about isolating the QR from the background. We should use fullScreenCover here I think?

GravatarApp/Share/ShareQRFullScreenView.swift
```diff
@@ -0,0 +1,102 @@
...
+                forceRefresh: $forceRefresh,
+                placeholderView: { EmptyView() }
+            )
+            .frame(maxHeight: .infinity)
+            .darkBlurImageStyle(width: windowWidth)
+        }
+        .environment(\.colorScheme, .dark)
+        .onAppear {
+            originalScreenBrightness = UIScreen.main.brightness
+            UIScreen.main.brightness = 1
```
pinarol, line 44: I felt like it's a bit too bright but I am not sure what the idea value would be

etoledom: This is a common pattern I see. And they use full brightness (I've checked 😅 )
We could try with 0.8 o 0.9 though. Since we expect this code to be read by other phones with relatively good cameras and no code reader machines which seem to be lower quality. 🤔 

Maybe it's not that important and we skip it for now?

GravatarApp/Share/ShareHeaderView.swift
etoledom: `fullScreenCover` looses the pull to dismiss gesture, but is needed for iPad to present full screen 🤔 

I went on using both 😅 
`sheet` on iPhone and `fullScreenCover` on iPad 👍

GravatarApp/Share/ShareQRFullScreenView.swift
pinarol: I see okay let's keep 1.

GravatarApp/Share/ShareHeaderView.swift
pinarol: Sounds good 👍

pinarol (approved):
Let's go :shipit: 🥳

PR #56, opened 2025-07-22
[Share screen] Share vcard action

etoledom:
Closes GRA-574

This PR implements sharing the contact vCard through system share sheet. It includes the user's avatar.

I've implemented a pre-loading of the avatar when the app launches, and every-time the user selects a different avatar.
This has an automatic re-try mechanism in case of failure, and a backup reload-if-needed when the view opens.

https://github.com/user-attachments/assets/74f3adaf-305d-49fd-b689-58bd094b2676

### To test:

- Try the share action button on the Share screen header menu.
  - Check that the avatar is properly added to any destination where the vcard is opened.

### DIFF
--- a/GravatarApp/Share/ShareViewModel.swift
+++ b/GravatarApp/Share/ShareViewModel.swift
@@ -12,10 +12,11 @@ class ShareViewModel: ObservableObject {
 
     private let userSession: UserSession
     private var cancellables = Set<AnyCancellable>()
-    private var avatarData: Data?
-    private var avatarDownloadTask: Task<Data?, Never>?
+    //   private var avatarData: Data?
+    //   private var avatarDownloadTask: Task<Data?, Never>?
     private let notificationCenter: NotificationCenter
     private let urlSession: any URLSessionProtocol
+    private let networkMonitor: NetworkMonitor
 
     let qrGenerator: QRGenerator
 
@@ -28,13 +29,15 @@ class ShareViewModel: ObservableObject {
     init(
         userSession: UserSession,
         notificationCenter: NotificationCenter = .default,
-        urlSession: URLSessionProtocol? = nil
+        urlSession: URLSessionProtocol? = nil,
+        networkMonitor: NetworkMonitor = SystemNetworkMonitor.shared
     ) {
         self.userSession = userSession
         self.profile = userSession.profile
         self.qrGenerator = QRGenerator()
         self.notificationCenter = notificationCenter
         self.urlSession = urlSession ?? GravatarURLSession.shared
+        self.networkMonitor = networkMonitor
 
         setupObservers()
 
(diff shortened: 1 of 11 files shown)


GravatarApp/Share/ShareViewModel.swift
```diff
@@ -51,22 +70,84 @@ class ShareViewModel: ObservableObject {
...
+                with: .hashID(profile.hash),
+                options: .init(preferredSize: .pixels(256))
+            )
+            return result.image.jpegData(compressionQuality: 0.7)
+        } catch {
+            if retryCount <= 20 {
+                try? await Task.sleep(for: .seconds(min(retryCount, 4)))
+                return await fetchAvatarData(retryCount: retryCount + 1)
+            }
+            return nil
```
pinarol, line 112: I mean this could work but I think we can have a more reliable solution if we combine the existing network monitor with the `forceRefreshAvatar` mechanism we already have. It's already used when switching the avatars. 

https://github.com/Automattic/Gravatar-iOS/pull/57/files 

The connectivity issues are the most common reasons for failure so I think we'll have a better chance with it.

After having these 2 mechanisms it would be really difficult to fail the avatar download imo but even so the worst thing that could happen would be to just not having the avatar in the contact card(or putting the previous one). And it's not something we should take this kind of extra measures I think.

GravatarApp/Share/ShareViewModel.swift
```diff
@@ -51,22 +70,84 @@ class ShareViewModel: ObservableObject {
...
+    }
+
+    private func getAvatarData() async -> Data? {
+        if let avatarData {
+            return avatarData
+        }
+        if let avatarDownloadTask {
+            return await avatarDownloadTask.value
+        }
+        return await fetchAvatarData()
```
pinarol: The underlying download service already has a synchronous memory cache so no need to retain and check the avatarData here imo.

GravatarApp/Share/ShareViewModel.swift
```diff
@@ -51,22 +70,84 @@ class ShareViewModel: ObservableObject {
...
+    }
+
+    private func fetchAvatarData(retryCount: Int = 0) async -> Data? {
+        let service = AvatarService(urlSession: urlSession)
+        do {
+            let result = try await service.fetch(
+                with: .hashID(profile.hash),
+                options: .init(preferredSize: .pixels(256))
```
pinarol: I see this size `.pixels(256)` everywhere, if we intend to download the same size I think we can declare it as a common constant somewhere. This would also help with sharing the same cache key underneath.

pinarol (commented):
Looks good thank you 🎉  Just want to propose these changes, I thought it'd be easier if I make a PR:

https://github.com/Automattic/Gravatar-iOS/pull/57/files

GravatarApp/Share/ShareViewModel.swift
etoledom: Thank you for the suggestions. Works great 🎉

etoledom:
@pinarol - ready for another look 👀

pinarol (approved):
Looks good 🎉

PR #57, opened 2025-07-22
Refresh avatar in QR tab

pinarol:
An alternative for the refresh mechanism https://github.com/Automattic/Gravatar-iOS/pull/56/


GravatarApp/AvatarPicker/AvatarPickerView.swift
```diff
@@ -64,6 +64,7 @@ struct AvatarPickerView: View {
                 MainMenuOptions(profile: avatarPickerModel.userSession.profile)
             } onRefresh: {
                 await avatarPickerModel.refresh()
+                avatarPickerModel.forceRefreshAvatar = true
```
pinarol, line 67: Noticed that we don't refresh the avatars during pull to refresh so I also added these.

PR #59, opened 2025-07-23
Adding sharing private info alert on Share screen

etoledom:
Closes GRA-573

This PR adds the alert explaining sharing private local information.

**Note:** Is not defined where the `Learn more` button should link to, so for now it does nothing.

### To test:

- Go to the Share screen
- Tap on the `!` symbol on the `Share private contact info.` section title.
  - Check that the alert is presented and looks as in the design.


pinarol (approved):
Looks great 🎉  We can remove the "Learn more" button as we discussed and :shipit: 🚀

PR #60, opened 2025-07-23
[Share screen] Status bar cover

etoledom:
Closes GRA-588
Closes GRA-587
Closes GRA-586

- Adding a small blurry view to cover the status bar when the Share screen scrolls.
- Adding a bit of extra top padding above the qr code image.
- Adding padding to the share screen when the top safe area is 0

https://github.com/user-attachments/assets/339c243f-9919-4aba-a041-4f8f28687e13

### To test
- Scroll the share screen
  - Check that the top cover only appears while scrolling.


GravatarApp/CommonViews/AnimatedHeaderScrollView/AnimatedHeaderScrollView.swift
```diff
@@ -30,7 +30,22 @@ struct AnimatedHeaderScrollView<ContentView: View, ScrollableHeader: View, Stick
     }
 
     private var scrollableHeaderTopPadding: CGFloat {
-        (safeAreaInset.top == 0 ? defaultTopPadding : safeAreaInset.top) + (isRefreshing ? 44 : 0)
+        (safeAreaInset.top == 0 ? defaultTopPadding : safeAreaInset.top) + (isRefreshing ? 44 : 0) + topPaddingCompensation
```
pinarol: I guess 44 is the refresh control's height? Could you instead use a common constant?

GravatarApp/CommonViews/AnimatedHeaderScrollView/AnimatedHeaderScrollView.swift
```diff
@@ -30,7 +30,22 @@ struct AnimatedHeaderScrollView<ContentView: View, ScrollableHeader: View, Stick
...
+    }
+
+    private var topPaddingCompensation: CGFloat {
+        safeAreaInset.top > 0 && safeAreaInset.top <= 20 ? 4 : 0
+    }
+
+    private let offsetReaderHeight: CGFloat = 8
```
pinarol, line 45: Since we moved the magic number here could you also move the code comment with it? The one that starts with "8 is a magic number found on testing. Not sure where is this being missed from..."

That's really useful.

pinarol (approved):
Looks good 🎉 Just adding 2 small comments. 
I don't think I'll need to check again.

We can tackle the faulty header that appears in iPad 13 inch 18.2 in another PR.

pinarol (approved):
iPad - iOS 18.2 problem seems solved. Thank you!

PR #61, opened 2025-07-23
Remove vision OS and landscape on iPhone

etoledom:
Closes GRA-589
Closes GRA-561

Remove vision OS and landscape on iPhone


pinarol (approved):
Looks good :shipit:

PR #62, opened 2025-07-23
Adding verified account to the Share screen and vcard

etoledom:
Closes GRA-572

This PR adds the verified account list to the share screen, and to the vcard shared via QR code and via share sheet.
I noticed we are not adding the Address (location) to the vcard yet, so I took this chance to also add it.

CleanShot 2025-07-23 at 14 29 48@2x

### To test:

- Log in with an account with one or more verified accounts
  - Check that the account fields appear in the Share screen 
  - Check that the switches selection persist over app relaunches
- Log in with an account without verified account
  - Check that no field with verified account is present.


GravatarApp/Share/ShareFields/ShareFieldsSelectionStore.swift
```diff
@@ -49,7 +52,18 @@ class ShareFieldsSelectionStore: ObservableObject {
...
 
-    init() {
+    func account(_ verifiedAcctoun: VerifiedAccount) -> Bool {
+        UserDefaults.standard.bool(forKey: verifiedAcctoun.url)
+    }
+
+    func set(_ verifiedAcctoun: VerifiedAccount, to value: Bool) {
+        UserDefaults.standard.set(value, forKey: verifiedAcctoun.url)
+        objectWillChange.send()
+    }
```
pinarol: The verified accounts have a different initial value(false) than others(true). I think all the fields should have the same initial value.

Also, the userDefault should better be injectable for unit testing. And using `@AppStorage` isn't helping us about it. This is not a View so it's not providing us auto-ui updates either. No need to use `@AppStorage` here.

I think we should just use the userDefaults directly here, solves all these things. I. made a PR for it: https://github.com/Automattic/Gravatar-iOS/pull/64

GravatarApp/Share/QRCode/VCardGenerator.swift
```diff
@@ -32,8 +47,16 @@ private func vCard(_ model: VCardModel) -> String {
     TEL:\(model.phoneNumber)
     EMAIL:\(model.email)
     URL:\(model.profileURL)
+    ADR;CHARSET=UTF-8;TYPE=HOME:;;;\(model.location);;;
```
pinarol: Good that you caught its absence. Although I noticed that the address field appears even if the location is empty 😬 It's tappable too. And it goes to Apple campus haha. Anyway I think we should remove this one if the location is empty?

Simulator Screenshot - iPhone 16 - 2025-07-23 at 21 26 32

etoledom: Yes, I didn't noticed that it shows up empty. We can remove the label if there's no location 👍

GravatarApp/Share/ShareFields/ShareFieldsSelectionStore.swift
etoledom: Yes I had this same feeling when I added user defaults for the verified accounts. I didn't put the time on it to think about how to implement the change.

Using user defaults directly totally looks like a better option. Thank you for working it out 🙏

pinarol (approved):
Let's go :shipit: 🥳

PR #63, opened 2025-07-23
Fix the QR size in iOS 18.2

pinarol:
The QR looks off in iOS 18.2.

simulator_screenshot_A85C6CE3-3AD7-47A6-8467-8B5516A727A2

This is because, to be able to calculate the windowWidth we need the QR component, but the QR component’s width also depends on windowWidth.

This PR calculates the window width on the parent level as a fix.

### DIFF
--- a/GravatarApp/Share/ShareViews/ShareHeaderView.swift
+++ b/GravatarApp/Share/ShareViews/ShareHeaderView.swift
@@ -75,9 +75,6 @@ struct ShareHeaderView<QRImage: View>: View {
             windowWidth: windowWidth,
             qrImage: qrImage
         )
-        .onChange(of: windowWidth) { _, newValue in
-            print("windowWidth: \(newValue)")
-        }
     }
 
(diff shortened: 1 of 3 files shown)


GravatarApp/Share/ShareViews/ShareHeaderView.swift
```diff
@@ -164,7 +168,8 @@ extension View {
                 geo.safeAreaInsets.top == 0 ?
                     16 : geo.safeAreaInsets.top,
                 imageURL: imageURL,
-                forceRefresh: .constant(false)
+                forceRefresh: .constant(false),
+                windowWidth: .constant(320)
```
etoledom: This value can be taken from the geometry reader instead of a constant:

```suggestion
                windowWidth: .constant(geo.size.width)
```

etoledom (approved):
Let's rock!

Great find 🎉 
:shipit:

PR #64, opened 2025-07-23
Refactor share fields selection store

pinarol:
Make `userDefaults` injectable.
Equalize all the default values to `true` including the verified accounts.


etoledom:
Great improvement! Thank you 🙏

PR #66, opened 2025-07-24
iPad improvements

etoledom:
Closes GRA-595

This PR does 4 things: 

- Center layout of the Share header on big screens
- Readable content width on the Welcome screen
- Fix the position of the close button on share QR full screen
- Use `UserDefaults` directly instead of `AppStorage` for Email and Phone number fields (needed for unit tests)

CleanShot 2025-07-24 at 10 54 19@2x
CleanShot 2025-07-24 at 10 54 23@2x
CleanShot 2025-07-24 at 10 54 32@2x

https://github.com/user-attachments/assets/9ae1d7e9-537f-4105-b199-99315da779e9

### To test

- Run the app on iPad
- Switch to the Share tab
- Test different widths with iPad split view
  - Check that the header adapts properly.
- Open the QR code full screen
  - Check that the close button is close to the QR code (nor far top)

*** Test also iOS18.2 and iPhone for regressions***


pinarol (approved):
Looks great :tada:

PR #67, opened 2025-07-24
Icons update

etoledom:
Closes GRA-597
Closes GRA-584

This PR updates some icons in the app.

- Using tab icons from the design
- Bundling the Photos picker button icon for iOS 17

Screenshot on iOS 17 with the Photos picker icon and the new tab bar icons:

CleanShot 2025-07-24 at 13 18 53@2x

### To test:

- Smoke test the tab icons


pinarol (approved):
Awesome, thank you 🎉

PR #70, opened 2025-07-24
Improve About modal view

etoledom:
Closes GRA-602

This PR improves the About modal view:

- Black action button
- Add links to TOS and Privacy Policy
- Layout closer to the design
- Added missing localized strings
- Code-wise: Grouping of sections

CleanShot 2025-07-24 at 22 43 17@2x

### To test:

- Check that TOS and PP links present the corresponding websites.


pinarol (commented):
We could show `Terms of service` and `Privacy policy` in an in-app safari I think, no need to leave the app for these 2 actions, wdyt?

cc: @AdamGrzybkowski @hamorillo

pinarol (approved):
Looks great 🎉 :shipit:

AdamGrzybkowski:
>We could show Terms of service and Privacy policy in an in-app safari I think, no need to leave the app for these 2 actions, wdyt?

We can change that. At the moment, we open it in the external browser.

pinarol:
> We can change that. At the moment, we open it in the external browser.

Yeah, let’s change it if it’s not too much trouble. Thanks! @AdamGrzybkowski

PR #71, opened 2025-07-25
[Share screen] Improve brightness control

etoledom:
Closes GRA-603

Manage the screen brightness in a way that we avoid increasing it to full brightness every time.

The lower the brightness value, we increase more, up until 0.5. From there, we increase a flat 0.2.

This comes from testing at night, where increasing 0.2 on a low brightness didn't feel enough.

### To test:

- Run the app on device.
- Go to the Share screen.
- Open the QR code full screen.
  - Check that the brightness increase feels right.


GravatarApp/Share/ShareViews/ShareQRFullScreenView.swift
```diff
@@ -53,6 +52,19 @@ struct ShareQRFullScreenView<QRImage: View>: View {
...
+        originalScreenBrightness = UIScreen.main.brightness
+        let increaseAmmount: CGFloat = {
+            guard originalScreenBrightness < 0.5 else {
+                return 0.2
+            }
+            // This returns 0.4 for 0.1 brightness, and 0.2 for 0.5 brightness.
+            // Then we continue increasing 0.2 afterwards.
+            return (0.9 - originalScreenBrightness) / 2
+        }()
+        UIScreen.main.brightness = min(originalScreenBrightness + increaseAmmount, 1)
```
pinarol, line 67: Works like a charm! 🥳

GravatarApp/Share/ShareViews/ShareQRFullScreenView.swift
```diff
@@ -43,8 +43,7 @@ struct ShareQRFullScreenView<QRImage: View>: View {
         }
         .environment(\.colorScheme, .dark)
         .onAppear {
-            originalScreenBrightness = UIScreen.main.brightness
-            UIScreen.main.brightness = 1
+            incrementBrightness()
         }
         .onDisappear {
             UIScreen.main.brightness = originalScreenBrightness
```
pinarol: One suggestion:

I got better results when I used this one instead of `onDisappear` when restoring the original brightness:

```
        .onChange(of: presentFullScreen) { oldValue, newValue in
            if !newValue {
                UIScreen.main.brightness = originalScreenBrightness
            }
        }
```

With `onDisappear`  the brightness lasts for an extra second on the parent page as well. It's not too bad but stands out especially in light mode. Ideally the brightness would go away on `willDisappear`, limiting it only to this page. We don't have an official `willDisappear` but we have the same effect with this one. WDYT?

etoledom: 💯 !

Feels much better 🎉 
I've pushed this change.

pinarol (approved):
Let's go! :shipit: Thank you for this adjustment! 🙇

PR #72, opened 2025-07-25
Remember the selected tab between app sessions

pinarol:
Closes GRA-585

## Description 

This PR persists the selected tab and restores it next time the app opens.

## To Test

Try selecting each tab, kill/reopen the app and check if:

- Selected tab is remembered
- There are no UI issues

### DIFF
--- a/GravatarApp/AppRoot/RootTabView.swift
+++ b/GravatarApp/AppRoot/RootTabView.swift
@@ -2,7 +2,7 @@ import Gravatar
 import SwiftData
 import SwiftUI
 
-enum RootTabItem: Int {
+fileprivate enum RootTabItem: Int {
     case gravatar = 0
     case profile
     case share
--- a/GravatarApp/AppRoot/RootTabView.swift
+++ b/GravatarApp/AppRoot/RootTabView.swift
@@ -41,7 +41,7 @@ struct RootTabView: View {
 
                 // MARK: - Second tab
 
-                ProfileTab(editProfileViewModel: editProfileViewModel)
+                ProfileTab(editProfileViewModel: editProfileViewModel, avatarForceRefresh: $avatarPickerViewModel.forceRefreshAvatar)
 
                 // MARK: - Third tab
 
@@ -78,10 +78,11 @@ struct GravatarTab: View {
 
 struct ProfileTab: View {
     @ObservedObject var editProfileViewModel: EditProfileViewModel
+    @Binding var avatarForceRefresh: Bool
 
     var body: some View {
         BackgroundColorView(color: .secondarySystemBackground) {
-            ProfileEditorView(viewModel: editProfileViewModel)
+            ProfileEditorView(viewModel: editProfileViewModel, forceRefreshAvatar: $avatarForceRefresh)
         }
         .tabItem {
             Label(Localized.profileTabTitle, image: .profileTab)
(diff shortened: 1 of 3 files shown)


GravatarApp/AppRoot/RootTabView.swift
```diff
@@ -2,12 +2,19 @@ import Gravatar
 import SwiftData
 import SwiftUI
 
+enum RootTabItem: Int {
+    case gravatar = 0
+    case profile
+    case share
+}
```
etoledom: This could be private. Or nested inside `RootTabView`. If it's nested it could be called `Tab`

etoledom (commented):
Works perfectly as described 🎉 

I did notice some weird behavior:
When the app launches on the Profile Editor, all other header avatars (from the other tabs) seem to reload from zero (without reading from cache). 

And also, the AvatarPicker avatars seem to re-load everytime the tab is selected.

This doesn't seem to happen when the app launches from Avatar Picker or Share screen.

https://github.com/user-attachments/assets/3ba92c28-b6cc-43aa-bf9b-3b31a8b7d9fa

etoledom:
> And also, the AvatarPicker avatars seem to re-load everytime the tab is selected.

I have double-checked with proximan.

The avatars are being force-refreshed everytime the AvatarPicker or the ShareScreen tabs are selected.
This only happens when the app starts from the ProfileEditor.

If the app starts from any other tab, there's no force refresh while selecting any of the tabs 🤔

GravatarApp/AppRoot/RootTabView.swift
pinarol: It's not actually private because it's used by the other Views in the same file, but I made it `fileprivate` for now.

GravatarApp/ProfileEditor/ProfileEditorView.swift
```diff
@@ -16,15 +15,15 @@ struct ProfileEditorView: View {
...
+                forceRefresh: $forceRefreshAvatar
             )
         } stickyHeader: { opacity, safeArea in
             ProfileEditorStickyHeaderView(
                 profile: viewModel.userSession.profile,
                 opacity: opacity,
                 safeAreaInsets: safeArea,
                 imageURL: headerAvatarURL,
-                forceRefresh: $forceRefresh
+                forceRefresh: $forceRefreshAvatar
```
pinarol, line 26: The profile editor didn't share the same force refresh flag so that was one problem.

GravatarApp/AvatarPicker/AvatarPickerViewModel.swift
```diff
@@ -90,8 +90,8 @@ class AvatarPickerViewModel: ObservableObject {
             .store(in: &cancellables)
 
         $selectedAvatarURL
-            .dropFirst()
             .removeDuplicates()
+            .dropFirst(2) // First value is nil, second value is the first time we load the URL, no need to `forceRefresh` for these cases.
```
pinarol, line 94: Another problem was the dropFirst() here, its order wasn't right and we should have passed 2 to it.

pinarol:
I believe I was able to fix that problem @etoledom

GravatarApp/AppRoot/RootTabView.swift
pinarol: Oh well swiftformat wasn't happy with fileprivate so I made it private.

etoledom (approved):
Working great and the problems are solved 🎉

GravatarApp/AppRoot/RootTabView.swift
etoledom: Yeah, `fileprivate` is used for fields inside a type to be accessible by other types in the same file.

But top level private declarations are still accessible in the same file, since they are declared top level.

PR #74, opened 2025-07-25
Update the icons on the avatar picker section

etoledom:
Closes GRA-607

This PR updates the icons on the avatar picker section, according to the designs.
This also fixes the different height between the buttons.

### To test:

- Smoke test the icons and size of the avatar picker buttons.


pinarol (approved):
Looks better this way, thank you! 🎉

PR #75, opened 2025-07-25
Remove unused buttons

pinarol:
Closes GRA-606

## Description

For now I just commented them out(should I have removed them instead?):
- [Avatar] Alt text 
- [Share header] Download

## To test

Check if the buttons are removed from the UI.


etoledom (approved):
Looking good!

I think is fine to keep them commented out. It will be easier to bring them back later on 👍

PR #76, opened 2025-07-25
Avoid sending tracks events to remote

etoledom:
Closes GRA-608

This PR adds a property to Analytics which allows to control if the track events will be send to remote.
For now, we set it to avoid events being tracked remotely.

Note: To see the Tracks log, you need to enable back `OS_ACTIVITY_MODE` on the schema Run arguments.

### To test:
- Activate `OS_ACTIVITY_MODE`
- Go throw the login flow.
- Send the app to the background
  -  Check that there are no logs from Tracks which send events to remote
  - You can double-check using Proxyman or similar software.
  - You can triple-check going to Tracks site to verify that there are no new track events from the app.

### DIFF
--- a/GravatarApp/AppRoot/GravatarAppApp.swift
+++ b/GravatarApp/AppRoot/GravatarAppApp.swift
@@ -1,7 +1,7 @@
+import Analytics
 import OAuth
 import SwiftData
 import SwiftUI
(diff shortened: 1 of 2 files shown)


pinarol (approved):
Looks good 🎉

PR #79, opened 2025-07-29
VCard update

etoledom:
Closes GRA-620

Updating the generated vcard format to work better between platforms.


pinarol (approved):
Looks good 🎉

PR #80, opened 2025-07-29
[Tooling] Add automation to generate strings from source

iangmaia:
Fixes AINFRA-1039

This PR adds a basic automation to generate strings based on the source code.
We can later plug this lane into more complex lanes and workflows (for example, during release management).
I've followed this approach also for better GlotPress compatibility.

I've also ran the lane so that the initial `en.lproj/Localizable.strings` is generated and I [imported](https://translate.wordpress.com/projects/gravatar/gravatar-ios/) it to GlotPress.


pinarol (approved):
Thank you 🙇

PR #81, opened 2025-07-29
Fix layout issue in iOS 17.5

etoledom:
Close GRA-632

I was able to reproduce this issue in iOS 17.4 and iOS 17.5. iPhone 14, 15, and iPad Pro M4.

I tested this solution in:

iOS: 17.4, 17.5, 18.2, 18.4, 18.5, 26 (xcode 16.4 build). iPhones and iPads

### To test:

- Check that the layout on the Share screen is correct.


pinarol (approved):
Tested with iPhone 17.5, 18.2, 18.5 and iPad 17.5, 18.2 18.5

All looks good 🎉

PR #82, opened 2025-07-30
Haptic feedback

etoledom:
Closes GRA-637

This PR adds some basic haptic feedback to the main interactions:

- Success toast (success feedback)
  - We do have `info` instead of `success`. But so far we only use it as success message. 
  - This triggers on:
    - Avatar selection success 
    - Profile save changes success
- Error toast (error feedback)
  - This triggers on:
    - Avatar selection error 
    - Profile save changes error
- Upload image success (success feedback)
- Upload image error (error feedback)
- Image picker buttons tap (selection feedback)
  - I added these since the buttons have little to no visual feedback, and sometimes they can be slow to react

### To test:

- Run the app on a device
- Try the previously described scenarios
  - Check that the haptic feedback triggers as expected and it feels OK

### DIFF
--- a/GravatarApp/AppRoot/RootTabView.swift
+++ b/GravatarApp/AppRoot/RootTabView.swift
@@ -58,6 +58,12 @@ struct RootTabView: View {
             }
         }
         .modalPresentation(manager: modalManager)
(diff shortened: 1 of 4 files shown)


pinarol (approved):
Very nice touch! Thank you! 🙇

PR #85, opened 2025-07-31
Add TestFlight Beta releases in CI

twstokes:
Fixes AINFRA-809

### Description

This PR contains the initial work for building and uploading beta releases to TestFlight on CI.

- Adds version management so that new builds are automatically bumped and synced with the Xcode project (four part semver)
- Kicks off a Buildkite pipeline to build the app and upload it to TestFlight

In the next iterations, we can further automate release uploading to other tracks (including production), updated metadata such as the app changelog, create GitHub releases, etc, while we also mature the release lifecycle according to the project's needs.

### Testing Steps

1. Make sure you have `rbenv` setup and environment variables such as `GITHUB_TOKEN` and `BUILDKITE_TOKEN`.
2. Checkout this branch (must be in a `release/*` branch)
3. Run `bundle install`
4. Run `bundle exec fastlane new_beta_release`. This will print what the lane will do and ask for confirmation.
5. After the confirmation, the lane should commit a version bump and trigger a build on CI. It should print the URL of the build so you can follow it.
6. The version should upload to TestFlight.


.buildkite/commands/release_upload.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -eu
+
+# Ensure we get the latest commit of the `release/*` branch, especially to get last version bump commit before publishing the GitHub Release and creating the git tag
+RELEASE_VERSION="${1:?RELEASE_VERSION parameter missing}"
+"$(dirname "${BASH_SOURCE[0]}")/checkout_release_branch.sh" "$RELEASE_VERSION"
+
+BETA_RELEASE=${2:-true} # use second call param, default to true for safety
+
+echo "Running $0 with BETA_RELEASE = $BETA_RELEASE..."
```
AliSoftware: This `BETA_RELEASE` env var doesn't seem to be used in the rest of the script.

I'm guessing the intent was to pass it to the `fastlane upload_app_store_connect_build_to_testflight beta_release:$BETA_RELEASE` call on line 24 at some point?

I get that this is an initial setup only and focused on beta at first, but I still think we should add the missing `beta_release:$BETA_RELEASE` param to the `fastlane` call just to not forget to pass it around when we'll work on prod releases in a later PR.

Even if the implementation of `upload_app_store_connect_build_to_testflight` have a `UI.user_error!('non-beta builds not implemented yet') unless beta_release` as its very first line  for now until we get to implementing the real support for it in a future PR, at least we won't risk to forget 😉

.buildkite/commands/release_upload.sh
```diff
@@ -0,0 +1,25 @@
...
+buildkite-agent artifact download "$ARTIFACTS_DIR/*.zip" . --step $STEP
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :closed_lock_with_key: Installing Secrets"
+bundle exec fastlane run configure_apply
+
+echo "--- :testflight: Uploading to TestFlight"
+bundle exec fastlane upload_app_store_connect_build_to_testflight
```
AliSoftware: 💄 Nit: I feel like "Upload App Store Connect Build" and "to TestFlight" might be a bit redundant?

Even when we will later implement uploads of production builds to App Store Connect and not just betas, the upload process still always goes through TestFlight anyway (the only difference being that for prod builds, after the upload to TestFlight is done, we do an extra step of selecting that TestFlight build for the version to be submitted).

So I think just naming the lane `upload_to_app_store_connect` would be less redundant and sufficient + simpler?

💡 For this current PR you'd just call that lane with `beta_release:true` always, and when the time will come to implement prod builds in later PRs you'll be able to use the same lane name (and just add the extra call to `upload_to_app_store_connect` for submission when `beta:false`), so one more reason to not include the term `testflight` in the lane name to begin with?

Config/Project.xcconfig
AliSoftware, line 1: 💭 I would suggest to split the `VERSION_LONG` and `VERSION_SHORT` entries into a dedicated `Version.xcconfig`

Not only for better separation of concerns (between settings that will be changed regularly vs setup that is not intended to change) but more importantly so that the version bump automation doesn't risk to accidentally delete some Xcode build settings from this `xcconfig` file when it re-writes it with new version values but might not ensure other build settings are also included in the re-write of the file.

---

[EDIT] After looking at [the implementation for `PUBLIC_VERSION_FILE.write(…)` in the `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/versioning/files/ios_version_file.rb#L56-L62) I see that it nicely reads the whole config file, then only replaces the settings for `VERSION_LONG`/`VERSION_SHORT`/`BUILD_NUMBER`, and writes the modified dict back. So it should actually preserve any other unrelated settings in the `.xcconfig` file it edits. Nice.

That being said, I still feel like it's more common in all our setups to have a dedicated file just for the versioning anyway, for better separation.

fastlane/Fastfile
```diff
@@ -119,6 +181,49 @@ platform :ios do
...
+    MESSAGE
+
+    Fastlane::Helper::GitHelper.checkout_and_pull(release: release_version_current) if is_ci?
+
+    UI.important(message)
+    UI.user_error!('Aborted by user request') unless skip_confirm || UI.confirm('Continue?')
+
+    UI.message 'Bumping build code...'
+    bump_build_code
+    UI.success "Done! New Build Code: #{build_code_current}"
```
AliSoftware: 💄 Nitpick: The term commonly used on iOS is "Build Number" rather than "Build Code", which is more an Android term.

For our helper methods in Release-Toolkit we settled on naming the methods of the `Wpmreleasetoolkit::Versioning::*` classes as `build_code` since those helpers are intended to be used in both iOS and Android and we had to settle on one naming, but for user-visible strings like the `UI.message` / `UI.success`, that will be read by iOS developers in logs, I'd suggest to use the "Build Number" terminology instead.

.buildkite/commands/release_upload.sh
twstokes: Good catch!

> I'm guessing the intent was to pass it to the fastlane upload_app_store_connect_build_to_testflight beta_release:$BETA_RELEASE call on line 24 at some point?

Most of the `.buildkite` work in this PR is lifted from [Pocket Casts iOS](https://github.com/Automattic/simplenote-ios/blob/b484bb016a0b788303da32091b0fea93ae51d6cc/.buildkite/commands/release-upload.sh) and when I removed the GitHub release I forgot to clean up this env var.

> upload_app_store_connect_build_to_testflight have a UI.user_error!('non-beta builds not implemented yet') unless beta_release

I can do this, but I'm trying to understand why that lane needs to differentiate between betas and prod releases in the first place? I thought that was more at the build finalization step, as well as a step after upload that could also submit it to the App Store for review (but not happen in this lane).

.buildkite/commands/release_upload.sh
twstokes: Yep, sounds good!

Config/Project.xcconfig
twstokes: Yep, can do.

fastlane/Fastfile
twstokes: Sounds good, will do.

.buildkite/commands/release_upload.sh
AliSoftware: > I can do this, but I'm trying to understand why that lane needs to differentiate between betas and prod releases in the first place? I thought that was more at the build finalization step, as well as a step after upload that could also submit it to the App Store for review (but not happen in this lane).

Yeah I can see both approaches:
 - Have a single lane `upload_to_app_store_connect do |beta_release:|` that calls `upload_to_testflight` in all cases, then only follow it with an `upload_to_app_store` call to submit it if `beta_release` is false
 - Have 2 lanes, `upload_build_to_testflight` that don't need a `beta_release:` parameter and just do that call to the `upload_to_testflight` action and nothing else, and a separate `submit_for_review` lane that calls the `upload_to_app_store` action, that would only be called by the pipeline if `BETA_RELEASE=true` / during the release finalization

I think both approaches are valid (though might be worth comparing how we might do it differently in PCiOS vs WPiOS/JPiOS/WCiOS/DOApple/… to find the most common approach we use in practice and prefer adopting the most common one for consistency… but not sure what it is just from the top of my head)

.buildkite/commands/release_upload.sh
twstokes: Should be resolved in a3fc7682a2eec88a4676ce3b98b1cb4cf8aa0749.

Config/Project.xcconfig
twstokes: >So it should actually preserve any other unrelated settings in the .xcconfig file it edits. Nice.

Comments aren't, so that'd be one reason to break this out to a dedicated version file.

Should be done in 9108feb69f1a131166af115db172c5642a49e37b.

fastlane/Fastfile
twstokes: Should be done in 722bf3d140bea6d0381e5ff10242c0112396d7e0.

AliSoftware: Seems you missed the UI.message from L210 😛

.buildkite/commands/release_upload.sh
twstokes: > and a `separate submit_for_review` lane that calls the `upload_to_app_store action`, that would only be called by the pipeline if BETA_RELEASE=true / during the release finalization

Just making sure, did you mean `BETA_RELEASE=false` here? Since beta releases would stick with TestFlight builds, but production (non-beta) releases would also be submitted to the App Store.

fastlane/Fastfile
twstokes: Oof. Thanks - eb7cf22bc186aa2cfced8581c03cd34850d71fe0.

.buildkite/commands/release_upload.sh
AliSoftware: Yep indeed (that was to check that you were following along 😇 😛  good catch!)

twstokes: I looked at the other apps and didn't spot a hard pattern for combing uploading and submitting to the App Store, so I propose we roll with what we have and I'll remove the unused env var for now.

Competed in 28645f0a79af15b5315a2d9479e249f83299fc5e.

twstokes:
Gonna go ahead and merge this one and follow up where needed.

PR #90, opened 2025-08-05
Fix caching issue on avatars

pinarol:
Closes GRA-655

## Description

Fixes an issue where the previous avatar remains on the header. We start to persist the previous avatar URL and compare it with the newly selected avatar URL. If they are different we start to clear the URLCache because otherwise it will still serve the older image. 

## Test

1. Select a different avatar, or remove the existing avatar (Try both)
2. Kill and reopen the app
3. Observe the correct avatar is visible on the headers

Repeat the test by doing step 1 from a different device/client.


pinarol:
Closing in favor of https://github.com/Automattic/Gravatar-iOS/pull/91

PR #95, opened 2025-08-07
Changes according to apple review

etoledom:
Closes GRA-697

On this PR:

- Re-introducing analytics tracking
- Adding extra tracking to welcome screen
- Remove extra animation when showing inline error messages
- Showing extra OAuth errors in an Alert
- Checking for internet connection before allowing to tap on the login button (with a no-internet message)
- Disabling the login button if there's no internet or in loading state.

### To test:

- Test the implementation of the previously described changes.

### DIFF
--- a/GravatarApp/Welcome/WelcomeViewModel.swift
+++ b/GravatarApp/Welcome/WelcomeViewModel.swift
@@ -175,7 +175,9 @@ class WelcomeViewModel: ObservableObject {
             await fetchProfile(with: token)
         } catch {
             analytics.track(WelcomeScreenEvent.oauthError(error: error.errorDescription))
-            if !(error.isAccessDenied || error.isCancelled) {
+            if error.isAssociatedDomainError {
+                oauthAlertErrorMessage = Localized.secureLoginErrorMessage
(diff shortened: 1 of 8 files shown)


GravatarApp/Welcome/WelcomeViewModel.swift
```diff
@@ -44,10 +61,15 @@ class WelcomeViewModel: ObservableObject {
         }
         do {
             isLoading = true
+            analytics.track(WelcomeScreenEvent.profileFetchStart)
             let profile = try await profileService.fetchOwnProfile(token: token)
+            analytics.track(WelcomeScreenEvent.profileFetchSuccess)
             configureSession(profile: profile, accessToken: token)
             cleanAllErrors()
         } catch {
+            if let error = error as? OAuthError { // Always the case (we need typed throws from the SDK)
```
pinarol: Confused. What throws `OAuthError` here?

etoledom: Oh good catch. It should be `APIError`

pinarol (approved):
Looks good ! :shipit:

PR #98, opened 2025-08-12
Add analytics for avatars screen

pinarol:
Closes GRA-718

Implements these events:

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-3b417869-7fff-2472-efc7-729fc598c0ac"><div dir="ltr" style="margin-left:0pt;" align="left">
Event name | Properties
-- | --
screen_view | screen: avatars
screen_leave | screen: avatars
mainmenu_tapped | screen: avatars
tab_gravatar_tapped |  
tab_profile_tapped |  
tab_qr_tapped |  
avatars_grid_item_tapped |  
avatars_camera_button_tapped |  
avatars_photos_button_tapped |  
avatars_ai_button_tapped |  

</div></b>

We apparently need a better way to detect taps on the `Menu` labels but that will be tackled on a separate PR. 

## To Test

Make sure this is unchecked to see the event logs in the XCode console. 

Screenshot 2025-08-13 at 16 44 00

Check the above events ☝️


GravatarApp/AvatarPicker/Avatar/AvatarActionsMenu.swift
```diff
@@ -11,23 +12,29 @@ struct AvatarActionsMenu<Label>: View where Label: View {
...
+                }
+
+                Section {
+                    button(for: .delete)
+                }
             }
-            Section {
-                button(for: .delete)
+            .onAppear {
+                labelTapAction?()
```
etoledom, line 37: As an option, we can get the analytics instance from environment on this view and track the action menu tap right here 🤔 

I'm not sure if it's better, but at least it will avoid the chain of callbacks up unitl `AvatarPickerView`

etoledom (approved):
Looking good!

One small observation in code.

We already talked about the issue tracking Menu taps, so I won't mention it here.

GravatarApp/AvatarPicker/Avatar/AvatarActionsMenu.swift
pinarol: Yeah it's really a close call for this one. Just that, it felt more consistent to do this way with the existing actions. Also keeping it decoupled from business logic at this child view level felt slightly more right.

PR #100, opened 2025-08-14
Using UIMenu instead of SwiftUI Menu

etoledom:
Closes GRA-727

On this PR we use UIMenu from UIKit on SwiftUI to be able to track menu taps.

#### About the implementation:

I tried to keep the SwiftUI style of declaration for the Menu items and sections, trying to avoid literal arrays as much as possible.

Since UIMenu uses UIActions as the menu item list, we can't use SwiftUI Buttons or any View, so I decided to declare MenuSection and MenuItem in a way that they can be used in a SwiftUI Views style, without being Views at all.

So we can have a UIKit menu declared as the follow, which feels like SwiftUI:

```swift
    MenuButton { // An actual SwiftUI View
        Text("Menu Button") // The menu button Label can be any SwiftUI View
    } content: {
        MenuItem(title: "Action 1") { }  // Not a SwiftUI View
        MenuItem(title: "Action 2") { }
        MenuItem(title: "Action 3") { }
    } onMenuAppear {
        print("Menu tapped!")
    }
```
With sections separator
```swift
    MenuButton {
        Text("Small Button") // The menu label can be any SwiftUI View
    } content: {
        // MenuItem and MenuSection are not `View`, and can only be used inside `MenuButton` content.
        MenuSection {
            MenuItem(title: "Action 1") { }
            MenuItem(title: "Action 2") { }
            MenuItem(title: "Action 3") { }
        }
        MenuSection {
            MenuItem(title: "Delete", systemImage: "trash", attributes: .destructive) { }
        }
    }
```

### To test:
- Tap on an avatar grid item
  - Check that the menu keeps working as usual
  - Check that all the menu items respond as expected
  - Check that the event `avatars_grid_item_tapped` event is tracked on every grid item tap.
- Tap on the Main Menu button on the Avatars View header,
  - Check that the menu keeps working as usual
  - Check that all the menu items respond as expected
  - Check that the event `mainmenu_tapped; properties: ["screen": AnyHashable("avatars")]` event is tracked on every grid item tap.


GravatarAppTests/AvatarTests/AvatarPickerViewSnapshotTests.swift
```diff
@@ -46,15 +49,12 @@ struct AvatarPickerViewSnapshotTests {
...
     @Test("AvatarPickerView displays correctly if the fetching happens when the grid is empty)", arguments: [true, false])
     func avatarFetchingWhenGirdIsEmpty(isSuccess: Bool) async throws {
-        let viewModel = AvatarPickerViewModel(
-            userSession: UserSession(profile: .testProfile, accessToken: "token", context: .testContext),
-            urlSession: URLSessionMock(returnErrorCode: isSuccess ? nil : 400)
-        )
+        urlSession.returnErrorCode = isSuccess ? nil : 400
         await viewModel.fetchAvatars()
         let view = AvatarPickerView(avatarPickerModel: viewModel, onLogout: {})
+            .environment(\.analytics, Analytics.test)
```
etoledom: We will need to be careful to not send the events to the prod tracker.

I don't think the test run long enough for the tracker to push the events to remote, but it could happen since they are stored persistently.

pinarol: Agreed. Although this can be easy to miss. Not for this PR, but maybe we can think about a more centralized mechanism to prevent analytics events from being sent during tests.

GravatarApp/CommonViews/MenuButton/MenuButton.swift
```diff
@@ -0,0 +1,94 @@
...
+
+    func makeUIView(context: Context) -> UIButton {
+        let button = UIButton(type: .system)
+        let hosting = UIHostingController(rootView: label())
+
+        addSubview(hosting.view, to: button)
+
+        button.menu = UIMenu(options: .displayInline, children: sections.map { $0.map() })
+        button.showsMenuAsPrimaryAction = true
+        button.addGestureRecognizer(UITapGestureRecognizer(target: context.coordinator, action: #selector(Coordinator.onMenuAppear)))
```
pinarol, line 29: aha nice!

pinarol (approved):
Looks pretty good 🎉 
Tested with iOS 17.5 and 18.6

PR #106, opened 2025-08-14
Analytics avatar actions

etoledom:
Closes: GRA-728

Adding analytics events for avatar actions:

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-f9b62548-7fff-f249-2edd-a914d3c829b5"><div dir="ltr" style="margin-left:0pt;" align="center">
Event name | Properties | Notes
-- | -- | --
avatar_action_select |   |  
avatar_action_share |   | Implemented in iOS
avatar_action_download |   | Implemented in Android
avatar_action_delete | is_selected: Bool |  
avatar_action_delete_warning_accepted | is_selected: Bool |  
avatar_action_delete_warning_cancelled | is_selected: Bool |  

</div><br /></b>

Also adding an event for uploading an image

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-44dfc15c-7fff-5721-231a-4c3fda0331ad"><div dir="ltr" style="margin-left:0pt;" align="left">
Event name | Properties | Notes
-- | -- | --
avatar_image_selected_for_upload | source: camera \| photo_library \| ai | Avatar upload intent, independent of upload success or error.

</div></b>

### To test:

- Check that the actions trigger the corresponding events


pinarol (commented):
Looks good! 🎉 

Just one thing, `avatar_image_to_upload_selected` is not very clear imo, how about `avatar_upload_image_selected`? And would be nice to add a `source` as an extra property there, could have values like:`camera | photo_library | playground`. WDYT?

etoledom:
> Looks good! 🎉
> 
> Just one thing, `avatar_image_to_upload_selected` is not very clear imo, how about `avatar_upload_image_selected`? And would be nice to add a `source` as an extra property there, could have values like:`camera | photo_library | playground`. WDYT?

This one is tricky, but I agree it's confusing.

I'd write `avatar_image_selected_to_upload`, but Tracks want the action at the end (unless is extra context, though `to_upload` might count as extra content actually)
So it's clear that an image has been selected and it is to be uploaded.
Should we use this one? @pinarol 

`avatar_upload_image_selected` could sound to me like an "upload image" button has been selected. Though I agree it's less confusing than the one currently used 🤔

etoledom:
@pinarol - Event name updated to `avatar_image_selected_for_upload`.
Also added the `source` property.

I've decided to use `ai` instead of `playground` to make it more multi-platform.

pinarol:
> I've decided to use ai instead of playground to make it more multi-platform.

I think the key point here is making sure we can clearly distinguish the source. If the specific image source is cross-platform then it's fine to give a cross-platform name. Playground is iOS only though, we know there won’t be a direct equivalent on Android. So it's more helpful to give it a distinguishable name. That way we know the usage for each tool separately. Not to mention also we can have multiple AI tools in the near future.

etoledom:
> Playground is iOS only though, we know there won’t be a direct equivalent on Android. So it's more helpful to give it a distinguishable name. That way we know the usage for each tool separately. Not to mention also we can have multiple AI tools in the near future.

In that case we can give the specific as a parameter?

We might want to know how many users use the AI tools in general. In the future when we do have many different AI tools which are needed to be tracked with different events, we can create those.

pinarol:
> In that case we can give the specific as a parameter?

> We might want to know how many users use the AI tools in general

Then we just sum the numbers, summing up is easy. `source` alone is sufficient for our case.

etoledom:
@pinarol - Updated event property name to `playground`

pinarol (approved):
Let's go 🚀

PR #107, opened 2025-08-14
Rename Share tab to QR

etoledom:
Closes GRA-732

This PR changes the Share tab name to QR

I did some research and I found out that the `QR` as a technical acronym, is universal and doesn't need localization.

- I also took the chance to create a make command to generate strings from source
  - `make generate-strings` and `make generate-strings-no-commit` 
- And I've run it to update the localized strings file.

CleanShot 2025-08-14 at 23 09 25@2x

### To test:

- Run te app
- Check the `QR` tab name


pinarol (approved):
LGTM 👍 
And thank you for the new makefile commands! 🙇

PR #108, opened 2025-08-15
Adding small animation on avatar tap

etoledom:
Closes GRA-666

This PR adds a small animation when an avatar in the grid is tapped. Since we don't have a `onMenuDisappear` event, we can't keep this state, so I decided to run the animation and finish it automatically.

This should help be sure of which avatar was tapped:

https://github.com/user-attachments/assets/d7826ece-3d41-4dd2-874f-14ec27892c2c

### To test:

- Tap on an avatar grid item
  - Check that the animation runs and it's smooth and helpful


pinarol (approved):
Looks good 🎉 Thank you!

PR #110, opened 2025-08-18
Integrating crash report with Privacy Settings

etoledom:
Closes GRA-714

This PR implements the Crash reports opt out option in Privacy Settings with Sentry Crash report.

I'll remove all "force crash" related code before merging.

### To test:

To force a crash:
- Open the MainMenu
- Tap on `Crash` option.

Test:
- With crash report sharing enabled by default:
  - Open the app
  - Force crash
  - Re-open the app
    - Check on Sentry that the new event has the ID of the user 
CleanShot 2025-08-18 at 14 37 00@2x

- Switch to no share crash reports
- Force crash
- Re-open the app
   - Check that there are no new events in Sentry
   
Test crash events is re-enabled without the need of restarting the app
- Enable sharing crash report
- Without re-starting the app, force crash 
- Re-open the app
  - Check that there's a new event on Sentry, but the amount of users affected didn't increase

This is right now (1 user is me). Probably with your tests, it should stay in 2.
CleanShot 2025-08-18 at 14 41 29@2x


pinarol (approved):
Looks great! :shipit:  🎉

PR #111, opened 2025-08-18
Create an app entry point for unit tests

etoledom:
Closes GRA-748

When Unit Tests run, the simulator launches running the app. This loads the Welcome Screen, and Welcome View Model, together with the Analytics and Crash reporting setup.

Since the Welcome Screen is shown, it will try to log a screen view event, independent of what tests are running, or if the analytics are mocked in the test suit.

With this PR we avoid all the overhead of loading the app and all its required components, instead it will show a text field which says "Unit Tests".

### To tests:

- Run the unit tests locally
- See the spawned simulator clone 
  - Check that the screen it shows is a black screen with the text "Unit Tests"
- Build the app for dev and for pord
  - Check that it build and run normally without errors.
  - (Prod build can be checked with the prototype build or choosing `release` in Run configuration)


pinarol (approved):
Looks good 👍

PR #114, opened 2025-08-19
Use sentry to log some errors

pinarol:
Closes GRA-751

## Description

Removing `oauth_error` and `profile_fetch_error` analytics events in favor of Sentry error logging.

Some interesting findings about Sentry:

1. I wanted to initially use `func logError(_ error: Error, userInfo: [String: Any]? = nil, level: SentryLevel = .error)` to put the already existing error description into the userInfo, but it gets removed due to scrubbing rules that we don't control.
 
Screenshot 2025-08-19 at 12 46 06

I also tried using the "tags" to add the error description but somehow it also doesn't work, the tag doesn't appear. Maybe there's a max length rule or something.

That said, Sentry already does a good job capturing the contents of the Error so hopefully we won't need the extra info.

2. I see only 1 row for different kinds of errors `oauth_error` and `profile_fetch_error`, so I added the tags to be able to query them easily. Tags enable us to do this:

Screenshot 2025-08-19 at 13 06 52
  
## To test

I used the RELEASE config since we don't send the logs to Sentry in DEBUG.

**Test 1 - `oauth_error`**

I enabled the Network Link Conditioner and set it to "Very bad network"
Removed the app from the sim
Reinstalled and quickly tapped to login
Observe that you see the "Setting up secure login… " alert

Wait for 15 seconds and restart
You should see this error being added to Sentry
Remember to disable the network link conditioner

**Test 2 - `profile_fetch_error`**

I used Proxyman's breakpoints to interrupt and abort the profile fetching service
Try logging in and simulate the profile fetch error
Wait for 15 seconds and restart
You should see this error being added to Sentry

### DIFF
--- a/GravatarApp/Extensions/APIError+Extensions.swift
+++ b/GravatarApp/Extensions/APIError+Extensions.swift
@@ -7,10 +7,10 @@ extension APIError {
         case .responseError(reason: .URLSessionError(error: let error)):
             let error = error as NSError
             return error.domain == NSURLErrorDomain &&
-            (
-                error.code == NSURLErrorNotConnectedToInternet ||
-                error.code == NSURLErrorNetworkConnectionLost
-            )
+                (
+                    error.code == NSURLErrorNotConnectedToInternet ||
+                        error.code == NSURLErrorNetworkConnectionLost
+                )
         default:
             return false
         }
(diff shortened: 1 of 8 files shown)


etoledom (approved):
Looking good!

:shipit:

PR #117, opened 2025-08-20
Marketing version fix

iangmaia:
Removing target-specific values for `MARKETING_VERSION`, which should use the value set in `GravatarApp.common.xcconfig` (that is, the generated `$(VERSION_SHORT)`).


AliSoftware (approved):

iangmaia:
cc @etoledom -- removing the target specific `MARKETING_VERSION` should fix the issue for the long run.

PR #118, opened 2025-08-20
Fix `AboutModalTests` unit tests

iangmaia:
The `AboutModalTests` tests seemed to have started failing after https://github.com/Automattic/Gravatar-iOS/pull/117.

That change made the Marketing Version to have an actual value in the tests target, and I believe reading `Bundle.main` won't have the right result (`Bundle.main` will refer to the test bundle, which doesn't have the same `Info.plist` configuration as the app bundle).

CI should be 🟢


pinarol (approved):
Looks good, thank you! 🎉

